PR #55, opened 2015-01-20
CB-8307 - new 10yr cert and tools to generate as needed

kirkshoop:
This change introduces a new certificate valid for 10 years.
The certificate will be updated for another 10 years on each 'npm install' (with no arguments)

The idea is to ensure that each published release contains a certificate that will be valid for 10 years from the release.

The keys are included because generating new keys is slow. Generating a new cert is fast.
The keys can be regenerated, and used to make a new certificate, with 'npm run-script generatekey'


sgrebnov:
:+1:

package.json
```diff
@@ -26,7 +28,8 @@
     },
     "devDependencies": {
         "jasmine-node": "1.14.5",
-        "rewire": ">=2.1.3"
+        "rewire": ">=2.1.3",
+        "node-forge": "~0.6.18"
```
sgrebnov: I believe `prepublish` is run during `install` so  all required deps must be in `dependencies` (not in `devDependencies`), otherwise `npm install --production` will fail.

kirkshoop: That is unfortunate. The goal is to only run before publishing a new release. installing from npm is not intended to run this. We really do not want to impose node-forge as a dependency. 
Do you have a recommendation? Is there any scenario where --production is required to work? If so, perhaps I can check to see if node-forge is installed from the script and skip the certificate creation.

kirkshoop:
@sgrebnov I changed the code to only use node-forge if it is installed. I tried `npm install --production` and the console reported that the certificate was not updated.

sgrebnov:
looks great!

purplecabbage:
I believe this still makes node-forge a dependency.
Running `npm install` will install all 'dependencies' and all 'devDependencies' and the only way to prevent it is to run `npm install --production`

kirkshoop:
@purplecabbage Yes, this does add a dependency on node-forge, but only for publishing the platform, not when using it. Thus the node-forge package has not been committed and the cert is only generated if node-forge is installed.
Is there a better way to accomplish this?

purplecabbage:
This may be better served living in cordova-coho which is the tool that we use to package a release. 
Or possibly have the generate_certificate.js call `npm install node-forge` so only people who call `npm generatekey` will ever need it.

I am not completely sure what the impact is of having node-forge as a dependency, it may not be that big of a deal, but we will be looking at some point to remove npm_modules folder and have platform installs actually use npm install directly. ie No bundling of the dependencies.

Not sure what is the best approach here.

PR #59, opened 2015-02-20
CB-8515 Support DefaultLanguage selection for Windows

muratsu:
Cordova should set the appropriate project file attributes for Windows 8.0, 8.1, and Windows Phone 8.1 using the W3C widget spec "defaultlocale" attribute. 

Related JIRA: https://issues.apache.org/jira/browse/CB-8515

### DIFF
--- a/template/cordova/lib/prepare.js
+++ b/template/cordova/lib/prepare.js
@@ -40,13 +40,34 @@ module.exports.applyPlatformConfig = function () {
         return false;
     });
 
+    // Apply appxmanifest changes
     ['package.windows.appxmanifest', 'package.windows80.appxmanifest', 'package.phone.appxmanifest'].forEach(
         function(manifestFile) {
             updateManifestFile(config, path.join(ROOT, manifestFile));
     });
 
+    // Apply jsproj changes
+    ['CordovaApp.Phone.jsproj', 'CordovaApp.Windows.jsproj', 'CordovaApp.Windows80.jsproj'].forEach(
+        function(jsprojFile) {
+            updatejsprojFile(config, path.join(ROOT, jsprojFile));
+    });
+
     copyImages(config);
 };
+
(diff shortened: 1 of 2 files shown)


template/cordova/lib/prepare.js
```diff
@@ -40,14 +40,37 @@ module.exports.applyPlatformConfig = function () {
...
+            updatejsprojFile(config, path.join(ROOT, jsprojFile));
+    });
+
     copyImages(config);
 };
 
+function updatejsprojFile(config, jsProjFilePath) {
+    var defaultLocale = config.defaultLocale();
+
+    if (defaultLocale) {
```
sgrebnov: I think we should remove or set to default DefaultLanguage in case there is no config.defaultLocale().

muratsu:
if no locale is not found it uses the current default language which is en-us. Since that is already hardcoded in proj files I'm not re-setting it. removing might have some side effects.

sgrebnov:
I meant if you changed locale in config.xml and then removed this settings it will continue pointing to the wrong previously used locale, not the default one.

sgrebnov:
May be just use something like this
var defaultLocale = config.defaultLocale() || 'en-us';

sgrebnov:
But this is minor, so I'm +1 merging this.

muratsu:
I'll fix the default locale to en-US. When a defaultlocale is not present we should set it to en-US. Good catch!

nikhilkh:
This looks good! @sgrebnov Can you please help merge this one? It will be good to have this in the next cordova-windows release.

cofetti:
This change is still not reflected in documentation in https://cordova.apache.org/docs/en/latest/config_ref/ (defaultlocale is shown as iOS-only)

janpio:
Documentation change is PRd, please have a look at https://github.com/apache/cordova-docs/pull/782

PR #64, opened 2015-03-27
CB-8760 'platform list' doesn't show version for windows platform.

TimBarham:
Added version file for `windows` platform (note that this relies on a change in `coho` to keep it in sync).

### DIFF
--- a/template/cordova/version
+++ b/template/cordova/version
@@ -1,0 +1,25 @@
+#!/usr/bin/env node
+
+/*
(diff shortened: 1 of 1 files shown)


TimBarham:
Hi - could someone take a look at this? A related change ([PR67](https://github.com/apache/cordova-coho/pull/67)) was merged, but this should really go in also. There's also another related change that needs to go in ([PR11](https://github.com/apache/cordova-browser/pull/11)). Thanks!

PR #65, opened 2015-04-09
CB-8307: Adding a 25-year expiration temporary certificate.

robpaveza:
Simplified: Just created a 25-year certificate.  We can investigate creating a certificate at platform-add or first-build time, but for now, this should be more than adequate to avoid build interruptions.


purplecabbage:
So we have 25 years to come up with a better solution? Sounds good to me.

robpaveza:
I've updated the certificate with a new EKU.  That was my bad for not verifying build with it.

For future reference (in 24.5 years when someone is digging through the history to get the correct commands...), the tools to generate the valid PFX for this are in the Windows SDK.  The correct usage is:

makecert -r -n "CN=Apache Cordova Team" -a sha256 -cy end -sky signature -eku 1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13 -sv CordovaApp_TemporaryKey.pvk CordovaApp_TemporaryKey.cer
pvk2pfx -pvk CordovaApp_TemporaryKey.pvk -spc CordovaApp_TemporaryKey.cer -pfx CordovaApp_TemporaryKey.pfx

The makecert parameters are documented on MSDN, but it comes down to self-signed, end-entity, signing, and the EKU parameters which I missed during the first commit (I thought makecert added that via the -sky signature parameter).  The EKU is for Code Signing and Lifetime Signing.

robpaveza:
Ping @purplecabbage  - can we get this merged in?  Thanks!

purplecabbage:
Done.

@purplecabbage
risingj.com

pke:
This somehow prevents me from specifying a publisherId myself for store submissions. I always get "CN=Apache Cordova Team". What do I have to do, to prevent that? `config.publisherId` set from build.json is ignored too.

PR #66, opened 2015-04-14
CB-8856 Fix 'Id' attr is invalid when creating Windows Store submission build

sgrebnov:
We should NOT change Application/Id in .appxmanifest according to widget/id (config.xml) as it does not make any sense, but could lead to build failures like in this particular case:

https://issues.apache.org/jira/browse/CB-8856


nikhilkh:
LGTM

PR #67, opened 2015-04-16
CB-8866: add Windows platforms "Mixed Platforms"

brodybits:
See [CB-8866](https://issues.apache.org/jira/browse/CB-8866): This is needed to install a plugin with a C++ component library such as: https://github.com/litehelpers/Cordova-sqlite-storage

If I would use the Cordova CLI to install the sqlite plugin, I can only build for certain CPU targets. This seems to be contrary to the "Windows Universal" approach.

Note my ICLA for Apache was already accepted for apache/cordova-wp8#72.

### DIFF
--- a/template/CordovaApp.sln
+++ b/template/CordovaApp.sln
@@ -41,10 +41,12 @@ Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Debug|ARM = Debug|ARM
+		Debug|Mixed Platforms = Debug|Mixed Platforms
 		Debug|x64 = Debug|x64
 		Debug|x86 = Debug|x86
 		Release|Any CPU = Release|Any CPU
 		Release|ARM = Release|ARM
+		Release|Mixed Platforms = Release|Mixed Platforms
 		Release|x64 = Release|x64
 		Release|x86 = Release|x86
 	EndGlobalSection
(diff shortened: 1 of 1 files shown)


brodybits:
When can this fix be considered? People are having problems using the sqlite plugin with the Visual Studio Cordova project template (ref: litehelpers/Cordova-sqlite-storage#237)

purplecabbage:
You cannot build for win32 with windows-universal. 
Universal in this case means wp8.1 and windows8.1 apps distributed by the windows app store.

brodybits:
I just pushed a new change to add "Mixed Platforms" *only*, based off the latest commit in the `master` branch. For reference purpose only, my first proposed fix is preserved in my `CB-8866-all` branch.

Please let me know if it is OK or not.

nikhilkh:
I believe cordova windows provides an --archs option to specify the architectures to build. Is that not sufficient for your purposes? Here's the exact syntax:
`cordova build windows -- --archs="x86"`

brodybits:
> I believe cordova windows provides an --archs option to specify the
> architectures to build. Is that not sufficient for your purposes? Here's the
> exact syntax:
> cordova build windows -- --archs="x86"

OK-then we need to select the architecture (`x86`|`x86-64`|`arm`) if we use Visual Studio as well.

Another thing is that people are having trouble using my plugin with a Visual Studio Cordova project, as reported in litehelpers/Cordova-sqlite-storage#237. Our workaround was to use `plugman`. I hope we can find a better solution for this.

These problems show up due to including a C++ project, which seems perfectly valid but not done for most Windows "Universal" plugins. I expect someone using https://github.com/AllJoyn-Cordova/cordova-plugin-alljoyn will encounter similar issues.

I am taking holiday this week, planning to try again to find a solution for the problem using the Visual Studio Cordova template when I come back.

robpaveza:
@brodybits: Have you been unblocked by using the `--archs` parameter?  

In my experience, building Windows apps using a "Mixed Platforms" solution platform is almost always undesirable.  Do you think we could close this PR out?

Thanks!

brodybits:
> Have you been unblocked by using the --archs parameter?

Yes. I think this should be documented very clearly (I will document this in my own plugin when I get a chance).

> In my experience, building Windows apps using a "Mixed Platforms" solution platform is almost always undesirable. Do you think we could close this PR out?

OK

PR #68, opened 2015-04-16
For discussion: CB-8869 remove Windows 8.0 from Windows Universal version

brodybits:
As I described [CB-8869](https://issues.apache.org/jira/browse/CB-8869), the Windows 8.0 target does not work with certain newer tools such as Visual Studio Express 2013, and is not supported by some newer Windows Universal plugins such as [Cordova-sqlite-storage](https://github.com/litehelpers/Cordova-sqlite-storage). Keeping the Windows 8.0 target can lead to confusion and the potential for other problems for new Windows Universal app developers.

If we cannot drop Windows 8.0 right now, then please ignore this PR.

As I said in PR #67, my ICLA for Apache was already accepted for apache/cordova-wp8#72.


brodybits:
Withdrawn. It is possible to configure an app to target Windows 8.1 in config.xml. To completely remove Windows 8.0 would entail removing CordovaApp.Windows80.jsproj, CordovaApp.vs2012.sln, and package.windows80.appxmanifest from the template then removing windows8.0 from the lib scripts and spec test scripts.

PR #74, opened 2015-05-05
CB-8946: Added the "WindowsToastCapable" preference 

robpaveza:
Added the "WindowsToastCapable" preference to indicate that the app can support toasts.  This is to support the Local Notifications plugin.  See https://issues.apache.org/jira/browse/CB-8946

### DIFF
--- a/template/cordova/lib/prepare.js
+++ b/template/cordova/lib/prepare.js
@@ -86,6 +86,7 @@ function updateManifestFile (config, manifestPath, namespacePrefix) {
     sortCapabilities(manifest);
     applyAccessRules(config, manifest);
     applyBackgroundColor(config, manifest, namespacePrefix);
+    applyToastCapability(config, manifest, namespacePrefix);
 
 
     //Write out manifest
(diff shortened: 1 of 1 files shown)


purplecabbage:
This should be done by the Local Notification plugin itself shouldn't it?
-1

robpaveza:
The plugin is likely unable to do this directly except via hooks, because the `<config-file>` directive is an element-level change, but the modification is at an attribute-level (Application/VisualElements@ToastCapable)

We have several preferences that modify the attributes of the VisualElements tag.  Plugins can specify elements become descendants of other elements in package.appxmanifest, but AFAIK they can't say "there's an existing element, set this attribute of that element" and have it respected by plugman.

Making such a change to plugman seems much larger and more invasive than creating a preference like this.

purplecabbage:
Okay, we can merge this.  I will file an issue to fix plugman also.
I would much rather if we could just go:
```<config-file target="package.phone.appxmanifest" parent="/Package/Applications/Application/m3:VisualElements[@ToastCapable]">true</config-file>
<config-file target="package.windows.appxmanifest" parent="/Package/Applications/Application/m3:VisualElements[@ToastCapable]">true</config-file>
<config-file target="package.windows80.appxmanifest" parent="/Package/Applications/Application/m3:VisualElements[@ToastCapable]">true</config-file>```

nikhilkh:
If you decide to use the `preference` approach - make sure you document it.

nikhilkh:
@robpaveza Has this been documented?

PR #75, opened 2015-05-05
CB-8486 Support for signing and build.json for Windows

muratsu:
Related JIRA item:
https://issues.apache.org/jira/browse/CB-8483

Details of the design and scenario here:
https://docs.google.com/document/d/1tJQ9OoGrrMhZcLI3mg46rGzAfbiQu9PuNBL1auAMGFM/edit#


template/cordova/lib/build.js
```diff
@@ -32,88 +33,175 @@ var projFiles = {
...
+    console.log('    --publisherId               : Sets publisher id field in manifest.');
+    console.log('    --buildConfig               : Sets build settings from configuration file.');
+    console.log('');
     console.log('examples:');
     console.log('    build ');
     console.log('    build --debug');
     console.log('    build --release');
     console.log('    build --release --archs="arm x86"');
+    console.log('    build --packageCertificateKeyFile="CordovaApp_TemporaryKey.pfx"');
+    console.log('    build --buildConfig="build.json"');
```
nikhilkh: As I mentioned earlier - providing an example with the publisher Id might be useful since publisher id has a very specific syntax.

nikhilkh:
Looks great, otherwise - let's get it in!

purplecabbage:
+1
and yeah, showing at least one valid publisher Id would be great.
I would recommend you also add a comment with a passing value whenever you write something like:
"(CN|L|O|OU|E|C|S|STREET|T|G|I|SN|DC|SERIALNUMBER|(OID\\.(0|[1-9][0-9]*)(\\.(0|[1-9][0-9]*))+))=" ...

nikhilkh:
LGTM! Let's get this in!

PR #77, opened 2015-05-14
CB-9025 Removes 'prepare' logic call as part of the 'build'.

vladimir-kotikov:
Fix for https://issues.apache.org/jira/browse/CB-9025


nikhilkh:
Please fix the jshint error and let's get this in today - before the next Cordova-windows release.

vladimir-kotikov:
@nikhilkh, @muratsu Please take a look again, since #75 has been merged before this, so i had to resolve conflict and update build config applying logic a bit. The main change is that `updateBuildConfig` is no more a part of `applyPlatformConfig` and thus called as a separate method during build only.

nikhilkh:
This has been merged as part of 53c5939. Please close the PR.

PR #78, opened 2015-05-19
CB-9045 Create a new solution for Win10/Dev15

muratsu:
Currently in dev12 solution we're using shared project. The shared project has some logic that doesn't work with windows10 projects. (This is due to version changes we have from 8, 8.1 to 10.X.Y.Z)

It looks like the best option we have is to seperate out win10 to a new solution.

### DIFF
--- a/template/CordovaApp.Windows10.jsproj
+++ b/template/CordovaApp.Windows10.jsproj
@@ -71,8 +71,15 @@
     <TargetPlatformVersion>10.0.10030.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.10030.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
-    <DefaultLanguage>en-US</DefaultLanguage>
-    <PackageCertificateKeyFile>CordovaApp_TemporaryKey.pfx</PackageCertificateKeyFile>
+  </PropertyGroup>
+  <!--
+    MSBuild\12.0\Bin\Microsoft.Common.CurrentVersion.targets contains a "TargetPlatformVersion" check, which will throw if
+    the version is not formatted as X.Y. To workaround this check, change target platform version to 10.0 if Visual Studio
+    Version is less than 14.
+  -->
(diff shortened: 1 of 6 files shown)


purplecabbage:
Is this the 'new' way, or is this a temporary step on the road to 10?

muratsu:
UAP and SharedProj dont play nice on VisualStudio13. My understanding is, in order to have backwards compat we need to split UAP from SharedProj. This provides a better user experience for both VS15 and VS13. Please note that this doesn't affect `cordova` commands much (very slight change in win10 jsproj) and is mostly for a better IDE experience.

purplecabbage:
Does it make sense to continue to support `cordova platform add windows` in this case? or should we add `cordova platform add windows10` ?
I for one prefer separation of concerns.  It seems to me we are jamming everything inside the `windows` platform which leads to multiple new cl flags so we can pull them back apart...

muratsu:
Hmm you're right on seperation of concerns. Having a new solution file isn't really a good solution to this problem. I think I have a workaround that will work with vs2013 - I'll send an update.

template/CordovaApp.Windows10.jsproj
```diff
@@ -86,12 +93,11 @@
     <Content Include="WinJS\js\WinJS.js" />
     <None Include="WinJS\js\WinJS.intellisense.js" />
     <None Include="WinJS\js\WinJS.intellisense-setup.js" />
-    <None Include="CordovaApp_TemporaryKey.pfx" />
   </ItemGroup>
-  <Import Project="CordovaApp.projitems" Label="Shared" />
+  <Import Project="CordovaApp.projitems" />
```
nikhilkh: Is removing `Label="Shared"` intentional?

nikhilkh:
LGTM. I prefer the revised version which ends up with only a single solution file. Consider squashing the two commits before merging.

template/CordovaApp.Windows10.jsproj
muratsu: I dont think we actually use this label but added back to make sure we dont break anything

PR #80, opened 2015-05-21
CB-9057: Replaced the Windows 10 Preview version of WinJS with the latest base.js

robpaveza:
Pulled the WinJS branch from github and built it then migrated to only base.js.  This removes the WinJS UI dependency and reduces build overhead.


purplecabbage:
Will this create a conflict if someone wants to use WinJS UI ?

robpaveza:
Sorry I commented on the Jira, not the PR:

No. Someone just drops the UI components (css/ui-light or dark, fonts/symbols.ttf, and ui.js) into the project.

You can get those components presently from git-cloning winjs from GitHub, and then doing a grunt build on the cloned repo.

nikhilkh:
I think the reference here: https://github.com/apache/cordova-windows/blob/master/cordova-js-src/platform.js needs to be updated to point to base.js

robpaveza:
Right you are.  I've rebased and updated the PR.

robpaveza:
Also I made a corresponding PR for Cordova.js.

https://github.com/apache/cordova-js/pull/113

peterjuras:
What is the recommended way to integrate ui.js now into a cordova app?

I don't want to include ui.js into every app version (iOS, Android & UWP) but just into UWP. Before this change cordova took care of that.

purplecabbage:
Sorry Peter, 
You are about 2 years too late. What is the goal you are trying to achieve?  
It is difficult to use different UI frameworks for different platforms ... although it IS possible through plugins.

peterjuras:
I'm trying to include ui.js only in the Windows build of our app. Do you know a plugin which could achieve this?

purplecabbage:
http://cordova.apache.org/docs/en/6.x/guide/cli/index.html#using-merges-to-customize-each-platform

PR #82, opened 2015-05-21
CB-9059 removed targetingWin10 logic from prepare

muratsu:
Removes isTargetingWin10 logic and fails silently in case system does not have UAP tools installed.


robpaveza:
LGTM

PR #83, opened 2015-05-22
CB-8954 Adds support for `requirements` command

vladimir-kotikov:
Implementation for https://issues.apache.org/jira/browse/CB-8954

### DIFF
--- a/bin/lib/check_reqs.js
+++ b/bin/lib/check_reqs.js
@@ -17,17 +17,110 @@
        under the License.
 */
 
-var Q     = require('Q'),
-    os    = require('os'),
+/*jshint node:true*/
+
+var Q     = require('Q');
+
+var MSBuildTools;
+try {
     MSBuildTools = require('../../template/cordova/lib/MSBuildTools');
+} catch (ex) {
+    // If previous import fails, we're probably running this script
+    // from installed platform and the module location is different.
+    MSBuildTools = require('./MSBuildTools');
+}
+
+/**
+ * Check if current OS is supports building windows platform
+ * @return {Promise} Promise either fullfilled or rejected with error message.
+ */
+var checkOS = function () {
+    var platform = process.platform;
+    return (platform === 'win32') ?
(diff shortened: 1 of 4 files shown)


bin/lib/check_reqs.js
```diff
@@ -17,17 +17,101 @@
...
+
+/**
+ * Checks if MSBuild tools is available.
+ * @return {Promise} Promise either fullfilled with MSBuild version
+ *                           or rejected with error message.
+ */
+var checkMSBuild = function () {
+    return MSBuildTools.findAvailableVersion()
+    .then(function (msbuildTools) {
+        // return Q.resolve('MSBuild tools v.' + msbuildTools.version + ' found at ' + msbuildTools.path);
```
TimBarham: Remove commented line.

bin/lib/check_reqs.js
```diff
@@ -17,17 +17,101 @@
...
+            requirement.installed = true;
+            requirement.metadata.version = version;
+        }, function (err) {
+            requirement.metadata.reason = err;
+        });
+    }, Q())
+    .then(function () {
+        // When chain is completed, return requirements array to upstream API
+        return requirements;
+    });
```
TimBarham: I know it's not a lot of code, but since it will be duplicated for every platform... it would be nice if the definition of `Requirement`, and the `check_all` logic could be in its own module that the platforms take as a dependency. A pain to set up now, but easier maintenance later on.

bin/lib/check_reqs.js
```diff
@@ -17,17 +17,101 @@
...
-    os    = require('os'),
+/*jshint node:true*/
+
+var Q     = require('Q');
+
+var MSBuildTools;
+try {
     MSBuildTools = require('../../template/cordova/lib/MSBuildTools');
+} catch (ex) {
+    // If previous import fails, we probably running this script
```
TimBarham: Typo: "we" should be "we're" or "we are".

PR #84, opened 2015-05-25
Draft implementation for Unified platform API

vladimir-kotikov:
Draft implementation of Unified Platform API for Windows platform, that inspired by [this thread](http://markmail.org/thread/3dw4mis4qo5d4ecz) in cordova-dev mailing list and [this proposal](https://github.com/kamrik/cordova-discuss/blob/master/proposals/PlatformProject.md)

Requires https://github.com/apache/cordova-lib/pull/226

### DIFF
--- a/template/cordova/Api.js
+++ b/template/cordova/Api.js
@@ -28,27 +28,12 @@ PlatformApi.prototype.requirements = function () {
     return requirementsImpl.check_all();
 };
 
-function PlatformHandler() {
-    // Set up basic properties. They probably will be overridden if this API is used by cordova-lib
-    this.root = path.join(__dirname, '..', '..');
-    this.platform = 'windows';
-
-    if (this.constructor.super_){
-        // This should only happen if this class is being instantiated from cordova-lib
-        // In this case the arguments is being passed from cordova-lib as well,
-        // so we don't need to care about whether they're correct ot not
-        this.constructor.super_.apply(this, arguments);
-    }
-}
-
-PlatformHandler.prototype.config_xml = function() {
+PlatformApi.prototype.getConfigXml = function() {
     return path.join(this.root, 'config.xml');
 };
 
-PlatformHandler.prototype.cordovajs_path = function() {
-    return path.join(this.root, 'www', 'cordova.js');
+PlatformApi.prototype.getWwwDir = function() {
+    return path.join(this.root, 'www');
 };
-
-PlatformApi.PlatformHandler = PlatformHandler;
 
 module.exports = PlatformApi;
--- a/template/cordova/Api.js
+++ b/template/cordova/Api.js
@@ -1,0 +1,54 @@
+
+/*jshint node: true*/
+
+var path = require('path');
+
+var buildImpl = require('./lib/build');
+var requirementsImpl = require('./lib/check_reqs');
+
+function PlatformApi () {
+    // Set up basic properties. They probably will be overridden if this API is used by cordova-lib
+    this.root = path.join(__dirname, '..', '..');
+    this.platform = 'windows';
+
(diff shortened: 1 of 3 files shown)


template/cordova/Api.js
```diff
@@ -0,0 +1,54 @@
...
+PlatformApi.prototype.build = function(context) {
+    var buildOptions = context && context.options || [];
+    return buildImpl.run(buildOptions);
+};
+
+PlatformApi.prototype.requirements = function () {
+    return requirementsImpl.check_all();
+};
+
+function PlatformHandler() {
```
kamrik: Do I understand correctly that we need to first move all/most of the code from cordova/metadata/windows_parser to this file before it gets fully operational?

kamrik: Do we want to keep PlatformHandler and PlatformApi as 2 separate classes long term?
From the current CLI style workflow, the methods in PlatformHandler seem to be like something private that should be hidden from the user. But actually some of the methods like getWwwDir() might be very useful to the user. An example would be when the web app is very large and copying it all on updateWww become a heavy operation. But if the user knows where to copy the files to, he can fine-tune his workflow to only copy changes which is a very common case when working with gulp/grunt watch.

vladimir-kotikov: > we need to first move all/most of the code from cordova/metadata/windows_parser to this file before it gets fully operational

This is correct. However, this shouldn't be a problem, because PlatformHandler interface is quite similar to platform parsers in cordova/metadata.

Regarding `PlatformHandler` and `PlatformApi` as 2 separate classes - I like your idea, but I would like to have a bit more feedback on this. The reason why it is implemented this way - i've tried to avoid changes that will break compatibility and obstruct the transition to new Api, so now they acts like former `PlatformProject` and its `parser` property.

BTW, another possible approach here, we're  thought about, was removing `parser` property at all and replace it and its methods with single `prepare()` methods which will act more like a "black box" - like `build()`, `run()` and others works now.

kamrik: The only reason I kept the parser property when switching to PlatformProject was backwards compatibility as well, with the hope to deprecate it once the rest of the code is changed to use methods of the PlatformProject directly. Actually there are very few direct calls to the parser left in cordova-lib, it's mostly used by the tests (which will have to change anyway).

The "prepare" has several sub-steps which might be interesting to the outside user as separate steps:
 * apply plugin config munges
 * apply config.xml based config changes to global project files
 * copy www
So I'm not sure prepare should be a single black box, but the above steps should definitely be methods of the PlatformApi rather than of some sub-object, and there can be a "prepare" method that executes all 3 of those steps for convenience.

Here is an example usage of a PlatformProject from a gulp file with separate calls to the above 3 stages.
https://github.com/kamrik/cordova-api-example/blob/master/gulpfile.js#L105
The plugin config munges are applied as part of gulp "create" task, and the other two as part of gulp "build" task. This allows for a much faster build for projects with many plugins. But slower for adding/removing plugins (which is a less frequent operation).

Whichever direction you decide to take it, thanks for picking up this API work!

vladimir-kotikov:
@kamrik Thanks a lot for your feedback, i've addressed some of your notes in https://github.com/MSOpenTech/cordova-windows/commit/6a27c27a569ee63894477d58d9593e4bad191e95 and https://github.com/MSOpenTech/cordova-lib/commit/039125e4280a8f3fa37f612aa78bf5d70ddaa220

kamrik:
Cool, looks much simpler now on the platform side and can let us eventually get rid of the handler when all platform move to use PlatformApi.

nikhilkh:
I think we should wait for the impending Cordova-windows 4.0 release before we merge this.

PR #85, opened 2015-05-26
CB-9073 Fixes build error when path to project contains `&` symbol

vladimir-kotikov:
This fixes [CB-9073](https://issues.apache.org/jira/browse/CB-9073)


purplecabbage:
Did you verify that paths with spaces are still allowed?

vladimir-kotikov:
Yup, this still works fine.

nikhilkh:
LGTM

PR #88, opened 2015-05-29
Update LICENSE & NOTE file to include appropriare third-party licenses

nikhilkh:


LICENSE
```diff
@@ -1,12 +1,309 @@
...
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+================================================================================
+node_modules/q
+================================================================================
+Copyright 2009�2012 Kristopher Michael Kowal. All rights reserved.
```
TimBarham: The hyphen is showing up as a weird character for me - perhaps should just replace with a normal hypen.

LICENSE
```diff
@@ -1,12 +1,309 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+ ADDITIONAL LICENSES:
+
+================================================================================
+/template/WinJS/js/base.js
+================================================================================
+
+WinJS
```
TimBarham: I think what was previously included in the NOTICE file would be sufficient here (that is, we don't need to include the full MIT license text - just mention that it is released under the MIT license and include the link - as we do below for `node-uuid`. No real harm in including the full text - it just makes the file larger and harder to follow.

LICENSE
```diff
@@ -1,12 +1,309 @@
...
+ARE DISCLAIMED. IN NO EVENT SHALL ARTUR ADIB BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+================================================================================
+node_modules/nopt
```
TimBarham: Where did this license text come from? It's different than the text in `node_modules/nopt/LICENSE`.

TimBarham:
What about `elementtree`? It's released under Apache license, but I think we still need to specifically mention it.

nikhilkh:
Updated. Thanks @TimBarham  for taking a look!

LICENSE
TimBarham: My bad. I had the wrong version of `nopt` installed.

PR #89, opened 2015-05-29
CB-9097 fail with a more descriptive error if run as admin

muratsu:
`cordova run` will fail for windows if invoked from admin prompt. Currently we will wait until ActivateApplication stage and fail afterwards. We should fail fast and improve user experience.


nikhilkh:
@vladimir-kotikov Could you please help review and test this one?

vladimir-kotikov:
LGTM

purplecabbage:
wfm! Merging.

PR #91, opened 2015-06-15
CB-9159 Fix Windows Phone 8.1 deploy when 'window-target-version' is 10

sgrebnov:
Jira issue: https://issues.apache.org/jira/browse/CB-9159

* Don't duplicate target processing logic in Run and Build scripts.
* Run should use first available build target returned by Build script
* Added additional tests for `--appx` parameter

### DIFF
--- a/template/cordova/lib/run.js
+++ b/template/cordova/lib/run.js
@@ -22,7 +22,6 @@ var Q = require('q'),
     path  = require('path'),
     build = require('./build'),
     utils = require('./utils'),
-    ConfigParser = require('./ConfigParser'),
     packages = require('./package'),
     execSync = require('child_process').execSync;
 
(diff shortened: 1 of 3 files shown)


sgrebnov:
@robpaveza could you please review?

nikhilkh:
@robpaveza Ping..

template/cordova/lib/run.js
```diff
@@ -169,20 +135,10 @@ module.exports.help = function () {
     process.exit(0);
 };
 
-
-function getWindowsTargetVersion() {
```
robpaveza: I'm confused about why this function is being removed.

robpaveza:
In general LGTM.  

I would prefer that we examine the pipeline here.  We had been talking about making the build process be module-based for Windows (for example, the Windows build step provides a set of require()-able modules that then executes from cordova-lib).  In such a world, we could just parse the CLI arguments once up-front, instead of making each individual step parse them, which seems to have increased the complexity as we talk about build vs run vs prepare.  I'd like to see us go that direction in the long-term, but in the short-term, I think this will be adequate.

sgrebnov:
I'll see if this is still actual and will do rebase on Monday.

template/cordova/lib/run.js
sgrebnov: This is not required anymore as we rely on build - we run the first build target.

var buildTargets = build.getBuildTargets(args.win, args.phone, args.appx);

..
+     if (!buildTargets || buildTargets.lenght <= 0) {
+         return Q.reject('Unable to determine deploy target.');
+     }
+
+     // we deploy the first build target so we use buildTargets[0] to determine
+     // what project type we should deploy
+     var projectType = projFileToType(buildTargets[0]);

vladimir-kotikov:
LGTM, merging.

PR #92, opened 2015-06-23
CB-9235 Adds more checks based on the windows-target-version

vladimir-kotikov:
This is an implementation for [CB-9235](https://issues.apache.org/jira/browse/CB-9235)

The logic of `check_all` method was updated to include additional checks for Visual Studio and  Windows/Windows phone SDKs.
Also logic of checks has been improved to checks for versions depending on the `windows-target-version` and `windows-phone-target-version` preferences from platform’s config.xml.

### DIFF
--- a/bin/lib/check_reqs.js
+++ b/bin/lib/check_reqs.js
@@ -68,12 +68,19 @@ var REQUIRED_VERSIONS = {
     }
 };
 
-var config = new ConfigParser(path.join(__dirname, '..', '..', 'config.xml'));
+function getConfig() {
+    try {
+        return new ConfigParser(path.join(__dirname, '..', '..', 'config.xml'));
+    } catch (e) {
+        throw new Error('Can\'t find config.xml file or it is malformed.');
+    }
+}
+
 
 function getMinimalRequiredVersionFor (requirement) {
 
-    var windowsTargetVersion = config.getWindowsTargetVersion();
-    var windowsPhoneTargetVersion = config.getWindowsPhoneTargetVersion();
+    var windowsTargetVersion = getConfig.getWindowsTargetVersion();
+    var windowsPhoneTargetVersion = getConfig.getWindowsPhoneTargetVersion();
     var windowsReqVersion = Version.tryParse(REQUIRED_VERSIONS[windowsTargetVersion][requirement]);
     var phoneReqVersion = Version.tryParse(REQUIRED_VERSIONS[windowsPhoneTargetVersion][requirement]);
 
@@ -259,7 +266,7 @@ var checkOS = function () {
         var requiredOsVersion = getMinimalRequiredVersionFor('os');
         if (actualVersion.gte(requiredOsVersion) ||
             // Special case for Windows 10/Phone 10  targets which can be built on Windows 7 (version 6.1)
-            actualVersion.major === 6 && actualVersion.minor === 1 && config.getWindowsTargetVersion() === '10.0') {
+            actualVersion.major === 6 && actualVersion.minor === 1 && getConfig.getWindowsTargetVersion() === '10.0') {
             return mapWindowsVersionToName(actualVersion);
         }
 
--- a/bin/lib/check_reqs.js
+++ b/bin/lib/check_reqs.js
@@ -19,15 +19,226 @@
 
 /*jshint node:true*/
 
-var Q     = require('Q');
+var Q     = require('q');
+var os    = require('os');
+var path  = require('path');
+var shell = require('shelljs');
 
-var MSBuildTools;
+var ConfigParser, MSBuildTools, Version, exec;
 try {
+    ConfigParser = require('../../template/cordova/lib/ConfigParser');
     MSBuildTools = require('../../template/cordova/lib/MSBuildTools');
+    exec = require('../../template/cordova/lib/exec');
+    Version = require('../../template/cordova/lib/Version');
 } catch (ex) {
     // If previous import fails, we're probably running this script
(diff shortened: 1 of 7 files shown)


nikhilkh:
Why are there AppVeyor failures

vladimir-kotikov:
I've slightly changed `Version` module to allow to accept incomplete version strings, but forgot ro check  test results :( However, if the test for incomplete strings is failing, this might be done intentionally. @robpaveza, could you please confirm that?

nikhilkh:
@robpaveza Can you please take a look?

bin/lib/check_reqs.js
```diff
@@ -19,15 +19,226 @@
...
+}
+
+/**
+ * Gets list of installed Windows Phone SDKs. Separately searches for 8.1 Phone
+ *   SDK and Windows 10 SDK, because the latter is needed for both Windows and
+ *   Windows Phone applications.
+ *
+ * @return  {Version[]}  List of installed Phone SDKs' versions.
+ */
+function getInstalledPhoneSdks () {
```
robpaveza: This doesn't match Windows 10, right?  This is explicitly only for WinPhone8.1?

bin/lib/check_reqs.js
```diff
@@ -19,15 +19,226 @@
...
+    },
+    '8.1': {
+        os: '6.2',
+        msbuild: '12.0',
+        visualstudio: '12.0',
+        windowssdk: '8.1',
+        phonesdk: '8.1'
+    },
+    '10.0': {
+        os: '6.2',
```
robpaveza: I think this is incorrect, I think it's actually 6.1 (Windows 7).

robpaveza: Also, Windows 8 is not supported for Win10 development.  You need Windows 8.1.  So it needs to be the set of (6.1, 6.3, 10.0).  I think you'll have to change your logic.

bin/lib/check_reqs.js
```diff
@@ -19,15 +19,226 @@
...
+        '6.2': 'Windows 8',
+        '6.3': 'Windows 8.1',
+        '10.0': 'Windows 10'
+    };
+    var majorMinor = shortenVersion(version);
+    return map[majorMinor];
+}
+
+function mapVSVersionToName(version) {
+    var map = {
```
robpaveza: You might suggest "2012 Express for Windows", "2013 Express for Windows", "2015 Community" to be specific.  Our documentation is incorrectly specific because it says "Professional."

bin/lib/check_reqs.js
vladimir-kotikov: No, this searches for both WP8.1 and Win10 SDKs - Win10 SDK is required wor building WP10 apps.

bin/lib/check_reqs.js
vladimir-kotikov: I've realized that there is another mistake from my side: there should be something like:
```javascript
'8.0': {
    os: '6.2', // instead of 6.1
    ...
},
'8.1': {
    os: '6.3', // instead of 6.2
    ...
},
'10.0': {
    os: '6.3', // Note that Windows 10 target is supported on Windows 7.
    ...        // This case get handled in checkOS function separately
    ...
```

bin/lib/check_reqs.js
vladimir-kotikov: Done.

robpaveza:
I'm checking out how it works with #102 .

robpaveza:
When I run this branch on my local, I get an npm test failure for check_reqs.spec.js, failing to find config.xml in cordova-windows root.

robpaveza:
build.spec.js spec.13 fails.

robpaveza:
I'm supposing build.spec.js spec.13 fails because it doesn't mock `findAllAvailableVersions` (or was also `findAvailableVersions` in your code).  I wrote a use case for this in #102 in spec.1; take a look at that and ensure the right code paths.  Otherwise, the merge of #92 and #102 is pretty straightforward.  Before merge, make sure your AppVeyor passes - I think it's a problem with check_reqs.

vladimir-kotikov:
> When I run this branch on my local, I get an npm test failure for check_reqs.spec.js, failing to find config.xml in cordova-windows root.

Sorry, that was my bad. [Fixed this here](https://github.com/MSOpenTech/cordova-windows/commit/e3f4a7880bbc9a5c3af0f09637235d66fe6762ba#diff-f477d7d5a4a7b8cc0bfe70b71a81920fR71)

> build.spec.js spec.13 fails because it doesn't mock findAllAvailableVersions

I can't repro this. build specs are all passing on my local machine. Probably merge issues?

> make sure your AppVeyor passes - I think it's a problem with check_reqs

AppVeyor tests was failing because AppVeyor machines has UAC disabled (I already mentioned that in one of my PRs, can't find it now), so the `run` command was failing. See https://github.com/apache/cordova-windows/blob/master/template/cordova/lib/run.js#L38. I've updated `run` command and its tests, so this condition is now mocked, and tests are passing even under the elevated prompt.

vladimir-kotikov:
> AppVeyor tests was failing because AppVeyor machines has UAC disabled (I already mentioned that in one of my PRs, can't find it now)

Found it: https://github.com/apache/cordova-windows/pull/99

sgrebnov:
lgtm :+1:

vladimir-kotikov:
This was merged in 9be50cd43a893f6af532b585baa7fa00e71772c9 but not auto-closed. Closing it now.

PR #93, opened 2015-06-25
CB-9252: Migrate WinJS to an NPM dependency

robpaveza:
This change puts the current WinJS NPM package into the node_modules
directory as well as package.json (for "npm install" support).  It then
modifies the create step to copy WinJS into the platform template
directory.


bin/lib/create.js
```diff
@@ -98,6 +106,16 @@ module.exports.run = function (argv) {
     return Q.resolve();
 };
 
+function recursiveCreateDirectory(targetPath) {
+    var parent = path.join(targetPath, '..');
+    if (!fs.existsSync(parent)) {
+        recursiveCreateDirectory(parent);
+    }
+
+    console.log('Creating directory "' + targetPath + '"');
```
TimBarham: Heh... Do we really want to be notified about each directory we're creating in the path? If we need any notification, I'd prefer just a single one (so at the call point rather than here).

PR #94, opened 2015-06-25
CB-8965 Copy cordova-js-src directory to platform folder during create

surajpindoria:


robpaveza:
Hey @surajpindoria - is this still a valid task?  It seems like cordova-lib is responsible for synthesizing cordova-js into a final file upon create.  What advantage do we get by having cordova-js-src into the packaged output instead of the built cordova.js file?

purplecabbage:
this is still valid. The src must be available for browserify and the exec/platform js code is tied to the template platform version.

PR #95, opened 2015-06-30
CB-9271: Removed the unnecessary device capabilities from the Windows 10

robpaveza:
Removed the unnecessary device capabilities from the Windows 10 app manifest.  These capabilities are no longer required as they no longer cause a crash in later flights of the OS.


nikhilkh:
LGTM

PR #96, opened 2015-07-07
CB-9283: Add support for Windows 10 WinAppDeployCmd

robpaveza:
The Windows 10 SDK uses a new "WinAppDeployCmd" utility for deployment to phones.  This change abstracts out the differences between Windows 8.1 and Windows 10 and the substantially different utility interfaces and factors that out into a separate deployment module.

### DIFF
--- a/spec/unit/deployment.spec.js
+++ b/spec/unit/deployment.spec.js
@@ -1,0 +1,282 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+var rewire     = require('rewire'),
+    deployment = rewire('../../template/cordova/lib/deployment'),
+    run        = deployment.__get__('run'),
+    Q          = require('q'),
+
+    AppDeployCmdTool = deployment.__get__('AppDeployCmdTool'),
+    WinAppDeployCmdTool = deployment.__get__('WinAppDeployCmdTool');
+
+var TEST_APP_PACKAGE_NAME = '"c:\\testapppackage.appx"',
+    TEST_APP_PACKAGE_ID   = '12121212-3434-3434-3434-567856785678';
+
+describe('The correct version of the app deployment tool is obtained.', function() {
+
+    it('Provides an AppDeployCmdTool when 8.1 is requested.', function() {
+
+        var tool = deployment.getDeploymentTool('8.1');
+        expect(tool instanceof AppDeployCmdTool).toBe(true);
+
+    });
+
+    it('Provides a WinAppDeployCmdTool when 10.0 is requested.', function() {
+
+        var tool = deployment.getDeploymentTool('10.0');
+        expect(tool instanceof WinAppDeployCmdTool).toBe(true);
+
+    });
+
+});
+
+describe('Windows 10 deployment interacts with the file system as expected.', function() {
+
+    function runMock(cmd, args, cwd) {
+        expect(cmd).toBe('c:\\Program Files (x86)\\Windows Kits\\10\\bin\\x86\\WinAppDeployCmd.exe');
+        switch (args[0]) {
+            case 'devices':
+                var output = 'Windows App Deployment Tool\r\nVersion 10.0.0.0\r\nCopyright (c) Microsoft Corporation. All rights reserved.\r\n\r\nDiscovering devices...\r\nIP Address      GUID                                    Model/Name\r\n127.0.0.1   00000015-b21e-0da9-0000-000000000000    Lumia 1520 (RM-940)\r\n10.120.70.172   00000000-0000-0000-0000-00155d619532    00155D619532\r\n10.120.68.150   00000000-0000-0000-0000-00155d011765    00155D011765\r\nDone.';
+                return Q(output);
+
+            case 'update':
+            case 'install':
+                expect(args[2]).toBe(TEST_APP_PACKAGE_NAME);
+                expect(args[4]).toBe('127.0.0.1');
+                return Q('');
+
+            case 'uninstall':
+                expect(args[2]).toBe(TEST_APP_PACKAGE_ID);
+                expect(args[4]).toBe('10.120.68.150');
+                return Q('');
+
+        }
+    }
+
+    var mockedProgramFiles = process.env['ProgramFiles(x86)'];
+    beforeEach(function() {
+        deployment.__set__('run', runMock);    
+        process.env['ProgramFiles(x86)'] = 'c:\\Program Files (x86)';
+    });
+    afterEach(function() {
+        deployment.__set__('run', run);
+        if (mockedProgramFiles) {
+            process.env['ProgramFiles(x86)'] = mockedProgramFiles;
+        } else {
+            delete process.env['ProgramFiles(x86)'];
+        }
+    });
+
+    it('enumerateDevices returns a valid set of objects', function() {
+        var deploymentTool = deployment.getDeploymentTool('10.0');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+
+            expect(deviceList.length).toBe(3);
+            expect(deviceList[0].name).toBe('Lumia 1520 (RM-940)');
+            expect(deviceList[0].index).toBe(0);
+            expect(deviceList[0].type).toBe('device');
+            
+            done = true;
+
+        });
+
+        waitsFor(function() { return done; });
+    });
+
+    it('installAppPackage passes the correct set of parameters', function() {
+        var deploymentTool = deployment.getDeploymentTool('10.0');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+            deploymentTool.installAppPackage(TEST_APP_PACKAGE_NAME, deviceList[0], /*shouldLaunch*/ false, /*shouldUpdate*/ false).then(function() {
+
+                // expect() calls are in the runMock function
+                done = true;
+
+            });
+        });
+
+        waitsFor(function() { return done; });
+    });
+
+    it('installAppPackage passes the correct set of parameters when updating', function() {
+        var deploymentTool = deployment.getDeploymentTool('10.0');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+            deploymentTool.installAppPackage(TEST_APP_PACKAGE_NAME, deviceList[0], /*shouldLaunch*/ false, /*shouldUpdate*/ true).then(function() {
+
+                // expect() calls are in the runMock function
+                done = true;
+
+            });
+        });
+
+        waitsFor(function() { return done; });
+    });
+
+    it('uninstallAppPackage passes the correct set of parameters', function() {
+        var deploymentTool = deployment.getDeploymentTool('10.0');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+            deploymentTool.uninstallAppPackage(TEST_APP_PACKAGE_ID, deviceList[2]).then(function() {
+
+                // expect() calls are in the runMock function
+                done = true;
+
+            });
+        });
+
+        waitsFor(function() { return done; });
+    });
+
+});
+
+describe('Windows 8.1 deployment interacts with the file system as expected.', function() {
+
+    function runMock(cmd, args, cwd) {
+        expect(cmd).toBe('c:\\Program Files (x86)\\Microsoft SDKs\\Windows Phone\\v8.1\\Tools\\AppDeploy\\AppDeployCmd.exe');
+        switch (args[0]) {
+            case '/EnumerateDevices':
+                var output = '\r\nDevice Index    Device Name\r\n------------    -------------------------------\r\n 0              Device\r\n 1              Mobile Emulator 10.0.10150.0 WVGA 4 inch 512MB\r\n 2              Mobile Emulator 10.0.10150.0 WVGA 4 inch 1GB\r\n 3              Mobile Emulator 10.0.10150.0 WXGA 4.5 inch 1GB\r\n 4              Mobile Emulator 10.0.10150.0 720p 5 inch 1GB\r\n 5              Mobile Emulator 10.0.10150.0 1080p 6 inch 2GB\r\n 6              Emulator 8.1 WVGA 4 inch 512MB\r\n 7              Emulator 8.1 WVGA 4 inch\r\n 8              Emulator 8.1 WXGA 4.5 inch\r\n 9              Emulator 8.1 720P 4.7 inch\r\n 10             Emulator 8.1 1080P 5.5 inch\r\n 11             Emulator 8.1 1080P 6 inch\r\nDone.\r\n';
+                return Q(output);
+
+            case '/update':
+            case '/install':
+            case '/updatelaunch':
+            case '/installlaunch':
+                expect(args[1]).toBe(TEST_APP_PACKAGE_NAME);
+                expect(args[2]).toBe('/targetdevice:de');
+                return Q('');
+
+            case '/uninstall':
+                expect(args[1]).toBe(TEST_APP_PACKAGE_ID);
+                expect(args[2]).toBe('/targetdevice:5');
+                return Q('');
+
+            default:
+                throw new Error('Unrecognized AppDeployCmd parameter "' + args[0] + '"');
+
+        }
+    }
+
+    var mockedProgramFiles = process.env['ProgramFiles(x86)'];
+    beforeEach(function() {
+        deployment.__set__('run', runMock);    
+        process.env['ProgramFiles(x86)'] = 'c:\\Program Files (x86)';
+    });
+    afterEach(function() {
+        deployment.__set__('run', run);
+        if (mockedProgramFiles) {
+            process.env['ProgramFiles(x86)'] = mockedProgramFiles;
+        } else {
+            delete process.env['ProgramFiles(x86)'];
+        }
+    });
+
+    it('enumerateDevices returns a valid set of objects', function() {
+        var deploymentTool = deployment.getDeploymentTool('8.1');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+
+            expect(deviceList.length).toBe(12);
+            expect(deviceList[0].name).toBe('Device');
+            expect(deviceList[0].index).toBe(0);
+            expect(deviceList[0].type).toBe('device');
+
+            expect(deviceList[5].name).toBe('Mobile Emulator 10.0.10150.0 1080p 6 inch 2GB');
+            expect(deviceList[5].index).toBe(5);
+            expect(deviceList[5].type).toBe('emulator');
+
+            done = true;
+
+        });
+        waitsFor(function() { return done; });
+    });
+
+    it('installAppPackage passes the correct set of parameters', function() {
+        var deploymentTool = deployment.getDeploymentTool('8.1');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+            deploymentTool.installAppPackage(TEST_APP_PACKAGE_NAME, deviceList[0], /*shouldLaunch*/ false, /*shouldUpdate*/ false).then(function() {
+
+                // expect() calls are in the runMock function
+                done = true;
+
+            });
+        });
+        waitsFor(function() { return done; });
+    });
+
+    it('installAppPackage passes the correct set of parameters when updating', function() {
+        var deploymentTool = deployment.getDeploymentTool('8.1');
+        var done = false;
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+            deploymentTool.installAppPackage(TEST_APP_PACKAGE_NAME, deviceList[0], /*shouldLaunch*/ false, /*shouldUpdate*/ true).then(function() {
+
+                // expect() calls are in the runMock function
+                done = true;
+
+            });
+        });
+        waitsFor(function() { return done; });
+    });
+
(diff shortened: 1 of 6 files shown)


template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+    DeploymentTool.call(this);
+    this.targetOsVersion = targetOsVersion;
+
+    /* jshint ignore:start */ /* Ignore jshint to use dot notation for 2nd process.env access for consistency */
+    var programFilesPath = process.env['ProgramFiles(x86)'] || process.env['ProgramFiles'];
+    this.path = path.join(programFilesPath, 'Microsoft SDKs', 'Windows Phone', 'v' + this.targetOsVersion, 'Tools', 'AppDeploy', 'AppDeployCmd.exe');
+    /* jshint ignore:end */
+}
+
+AppDeployCmdTool.prototype = new DeploymentTool();
```
nikhilkh: This link seems to suggest you should use `Object.create` : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Introduction_to_Object-Oriented_JavaScript

robpaveza: The only reason to prefer "Object.create(T)" over "new T()" is the case in which the T function is parameterized.  However, in this case it isn't; and callers referring to the deployment module can't see that anyway.

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+var Q     = require('q'),
+    fs    = require('fs'),
+    /* jshint ignore:start */ // 'path' only used in ignored blocks 
+    path  = require('path'),
+    /* jshint ignore:end */
+    proc  = require('child_process');
+
+// neither 'exec' nor 'spawn' was sufficient because we need to pass arguments via spawn
+// but also need to be able to capture stdout / stderr
+function run(cmd, args, opt_cwd) {
```
muratsu: I see this function everywhere in cordova codebase. We should really make a npm module out of it and publish it.
Better(?) version of this function is available at cordova-lib\cordova-lib\src\cordova\superspawn.js

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+       software distributed under the License is distributed on an
+       "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+       KIND, either express or implied.  See the License for the
+       specific language governing permissions and limitations
+       under the License.
+*/
+
+var Q     = require('q'),
+    fs    = require('fs'),
+    /* jshint ignore:start */ // 'path' only used in ignored blocks
```
muratsu: nit: trailing whitespace

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Determines whether the requested version of the deployment tool is available.
+ * @returns True if the deployment tool can function; false if not.
+ */
+DeploymentTool.prototype.isAvailable = function() {
+    return fs.existsSync(this.path);
+};
+
+/**
```
muratsu: nit: trailing whitespace

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+ * @returns A Promise which will be fulfilled on success or rejected on failure.
+ * @param packageInfo The app package name or Phone GUID representing the app.
+ * @param targetDevice An object returned from a successful call to enumerateDevices.
+ */
+DeploymentTool.prototype.uninstallAppPackage = function(packageInfo, targetDevice) {
+    return Q.reject('Unable to uninstall any app packages because that feature is not supported.');
+};
+
+/**
+ * Gets a list of installed apps on the target device.  This function is not supported for
```
muratsu: nit: trailing whitespace

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+ * @param packageInfo {String} The app package name or Phone GUID representing the app.
+ * @param targetDevice {Object} An object returned from a successful call to enumerateDevices.
+ * @returns A Promise for when the app is launched.
+ */
+DeploymentTool.prototype.launchApp = function(packageInfo, targetDevice) {
+    return Q.reject('Unable to launch an app because that feature is not supported.');
+};
+
+/**
+ * Gets a DeploymentTool for use in
```
muratsu: nit: trailing whitespace

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+        var lines = result.split('\n');
+        var matchedLines = lines.filter(function(line) {
+            return LINE_TEST.test(line);
+        });
+
+        var devices = matchedLines.map(function(line, arrayIndex) {
+            var match = line.match(LINE_TEST);
+            var index = parseInt(match[1], 10);
+            var name = match[2];
+
```
muratsu: nit: trailing whitespace

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+AppDeployCmdTool.prototype.uninstallAppPackage = function(packageInfo, targetDevice) {
+    return run(this.path, ['/uninstall', packageInfo, '/targetdevice:' + targetDevice.__shorthand]);
+};
+
+AppDeployCmdTool.prototype.launchApp = function(packageInfo, targetDevice) {
+    return run(this.path, ['/launch', packageInfo, '/targetdevice:' + targetDevice.__shorthand]);
+};
+
+function WinAppDeployCmdTool(targetOsVersion) {
+    if (!(this instanceof WinAppDeployCmdTool))
```
muratsu: nit: trailing whitespace

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+
+DeviceInfo.prototype.toString = function() {
+    return this.index + '. ' + this.name + ' (' + this.type + ')';
+};
+
+function AppDeployCmdTool(targetOsVersion) {
+    if (!(this instanceof AppDeployCmdTool))
+        throw new ReferenceError('Only create an AppDeployCmdTool as an instance object.');
+
+    DeploymentTool.call(this);
```
muratsu: Is this call required?

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+ */
+DeploymentTool.prototype.launchApp = function(packageInfo, targetDevice) {
+    return Q.reject('Unable to launch an app because that feature is not supported.');
+};
+
+/**
+ * Gets a DeploymentTool for use in 
+ */
+DeploymentTool.getDeploymentTool = function(targetOsVersion) {
+    if (targetOsVersion === '8.0' || targetOsVersion === '8.1') {
```
muratsu: Target OS version 8.0 is not supported.

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+        var devices = matchedLines.map(function(line, arrayIndex) {
+            var match = line.match(LINE_TEST);
+            var ip = match[1];
+            var guid = match[2];
+            var name = match[3];
+            var type = 'device';
+
+            var deviceInfo = new DeviceInfo(arrayIndex, name, type);
+            deviceInfo.__ip = ip;
+            deviceInfo.__guid = guid;
```
muratsu: __guid seems like unused

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,272 @@
...
+            return deviceInfo;
+        });
+
+        return devices;
+    });
+};
+
+WinAppDeployCmdTool.prototype.installAppPackage = function(pathToAppxPackage, targetDevice, shouldLaunch, shouldUpdate, pin) {
+    if (shouldLaunch) {
+        console.warn('Warning: Cannot launch deployed app with current version of Windows 10 SDK tools.');
```
muratsu: nit: terminal message is longer than 80 chars

template/cordova/lib/target-list.js
```diff
@@ -17,20 +17,21 @@
...
+    console.log('    --win10             : Chooses to list Windows 10 devices (Windows 8.1 is default).');
     console.log('    --emulators         : List the possible target emulators availible.');
-    console.log('    --devices           : List the possible target devices availible. *NOT IMPLEMENTED YET*');
+    console.log('    --devices           : List the possible target devices availible.');
     console.log('    --started_emulators : List any started emulators availible. *NOT IMPLEMENTED YET*');
     console.log('    --all               : List all available devices');
     console.log('examples:');
-    console.log('    node target-list.js --emulators');
-    console.log('    node target-list.js --devices');
+    console.log('    node target-list.js --win8.1 --emulators');
```
muratsu: --win8.1 flag is unhandled

template/cordova/lib/target-list.js
```diff
@@ -40,10 +41,42 @@ function help() {
...
+    if (args.indexOf('--emulators') >= 0) {
+        onlyEmulators = true;
+    }
+
+    if (onlyDevices && onlyEmulators) {
+        console.warn('Error: Cannot specify both --emulators and --devices');
+        help();
+        return;
+    }
+
```
muratsu: nit: trailing whitespace

muratsu:
Can you check template/cordova/lib/list-devices.bat and it's siblings to see if they still work as expected? This is a big refactor.

template/cordova/lib/deployment.js
robpaveza: You were so excited to nit the trailing whitespace you didn't note that the comment was incomplete.

template/cordova/lib/deployment.js
robpaveza: No, but it's convention because AppDeployCmdTool "derives from" DeploymentTool.  The fact that DeploymentTool doesn't do anything in its constructor would make eliminating this a micro-optimization, but is unnecessary.

template/cordova/lib/deployment.js
robpaveza: True.  We could create a separate jira (I don't want to dot out of cordova-windows to access superspawn here) to do this refactor.

template/cordova/lib/deployment.js
robpaveza: I believe guid will be interesting in a future version of cordova-windows, so I'd like to preserve it here.  The information is cheap to retain.

template/cordova/lib/target-list.js
robpaveza: Good catch, it had been dropped.

robpaveza:
The siblings work as expected, although they only support Windows 8.1.  We'll have to make a decision about how to invoke the Windows 10 equivalent.

spec/unit/deployment.spec.js
```diff
@@ -0,0 +1,282 @@
...
+        deploymentTool.enumerateDevices().then(function(deviceList) {
+            deploymentTool.installAppPackage(TEST_APP_PACKAGE_NAME, deviceList[0], false, false).then(function() {
+
+                // expect() calls are in the runMock function
+                done = true;
+
+            });
+        });
+
+        waitsFor(function() { return done; });
```
nikhilkh: Is this the old way of doing async on Jasmine?. We should really upgrade to newer version of Jasmine as it offers better async support.

template/cordova/lib/deployment.js
nikhilkh: @vladimir-kotikov is working on cordova-common module that all platforms and cordova-lib could take a dependency on. At that point supersawn could be one of those things we could share.

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,275 @@
...
+        child.stderr.on('data', function(s) { stderr += s; });
+        child.on('exit', function(code) {
+            if (code) {
+                d.reject(stderr);
+            } else {
+                d.resolve(stdout);
+            }
+        });
+    } catch(e) {
+        console.error('error caught: ' + e);
```
nikhilkh: Should this method be responsible for logging error or the caller?

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,275 @@
...
+
+        var devices = matchedLines.map(function(line, arrayIndex) {
+            var match = line.match(LINE_TEST);
+            var ip = match[1];
+            var guid = match[2];
+            var name = match[3];
+            var type = 'device';
+
+            var deviceInfo = new DeviceInfo(arrayIndex, name, type);
+            deviceInfo.__ip = ip;
```
nikhilkh: I'm curious - why use `__` - its hard to read. I have not seen this convention before.

template/cordova/lib/package.js
```diff
@@ -130,58 +145,47 @@ module.exports.findDevice = function (target) {
...
+    return deployment.then(function(deploymentTool) {
+        return module.exports.findDevice(deploymentTool, deployTarget).then(function(target) {
+            return uninstallAppFromPhone(deploymentTool, package, target).then(
                 function() {}, function() {}).then(function() {
-                    return installAppToPhone(appDeployUtils, package, target, false);
-                }).then(function() {
-                    return runAppOnPhone(appDeployUtils, target);
-                }, function(error) {
+                    // shouldUpdate = false because we've already uninstalled
+                    return deploymentTool.installAppPackage(package.appx, target, true, false);
```
nikhilkh: It helps readability to have simple comments /*shouldLaunch=*/ true.

template/cordova/lib/package.js
```diff
@@ -130,58 +145,47 @@ module.exports.findDevice = function (target) {
...
+                }).then(function() { }, function(error) {
+                    console.dir(error);
```
nikhilkh: should the console.dir be removed?

template/cordova/lib/run.js
```diff
@@ -155,7 +155,8 @@ module.exports.help = function () {
     console.log('    --appx=<8.1-win|8.1-phone|uap>');
     console.log('                  : Overrides windows-target-version to build Windows 8.1, ');
     console.log('                              Windows Phone 8.1, or Windows 10.');
-    console.log('');
+    console.log('    --win10tools  : Uses Windows 10 deployment tools (used for a Windows 8.1 app when');
```
nikhilkh: Why do we need to expose this as an option to the user?

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,275 @@
...
+    this.path = path.join(programFilesPath, 'Windows Kits', '10', 'bin', 'x86', 'WinAppDeployCmd.exe');
+    /* jshint ignore:end */
+}
+
+WinAppDeployCmdTool.prototype = Object.create(WinAppDeployCmdTool);
+WinAppDeployCmdTool.prototype.constructor = WinAppDeployCmdTool;
+
+WinAppDeployCmdTool.prototype.enumerateDevices = function() {
+    var that = this;
+    var LINE_TEST = /^([\d\.]+?)\s+([\da-fA-F\-]+?)\s+(.+)$/m;
```
nikhilkh: An example string comment really helps understand regex usage.

template/cordova/lib/package.js
```diff
@@ -113,16 +112,32 @@ function getPackageName(platformPath) {
...
-        for (var idx in sortedList){
-            if (sortedList[idx].toLowerCase().indexOf(target) > -1) {
-                // we should return index based on original list
-                return Q.resolve(deviceList.indexOf(sortedList[idx]));
+        // In CB-9283, we need to differentiate between emulator, device, and target.
+        // So, for emulators to honor the above CB-7617, we preserve the original behavior.
+        // Else, we choose either the target by ID (DeviceInfo.index) or if it's just device,
+        // we choose the default (aka first) device.
+        if (target === 'emulator') {
+            var sortedList = deviceList.concat().sort(function (l, r) { return l.toString().length > r.toString().length; });
```
nikhilkh: I know this is existing code - but why would you `concat` here to nothing.

template/cordova/lib/package.js
```diff
@@ -113,16 +112,32 @@ function getPackageName(platformPath) {
...
+            for (var idx in sortedList){
```
nikhilkh: If I understand correctly, `sortedList` is an array and foreach is quite a bad practice for arrays - unless we really have a case of a sparse array.

template/cordova/lib/deployment.js
```diff
@@ -0,0 +1,275 @@
...
+DeploymentTool.prototype.launchApp = function(packageInfo, targetDevice) {
+    return Q.reject('Unable to launch an app because that feature is not supported.');
+};
+
+/**
+ * Gets a DeploymentTool to deploy to devices or emulators.
+ * @param targetOsVersion {String} The version of the 
+ */
+DeploymentTool.getDeploymentTool = function(targetOsVersion) {
+    if (targetOsVersion === '8.1') {
```
nikhilkh: It will be nice to have this as an enum rather than literal strings.

template/cordova/lib/target-list.js
```diff
@@ -17,20 +17,21 @@
...
+    console.log('    --win10             : Chooses to list Windows 10 devices (Windows 8.1 is default).');
     console.log('    --emulators         : List the possible target emulators availible.');
-    console.log('    --devices           : List the possible target devices availible. *NOT IMPLEMENTED YET*');
+    console.log('    --devices           : List the possible target devices availible.');
     console.log('    --started_emulators : List any started emulators availible. *NOT IMPLEMENTED YET*');
     console.log('    --all               : List all available devices');
     console.log('examples:');
     console.log('    node target-list.js --emulators');
-    console.log('    node target-list.js --devices');
+    console.log('    node target-list.js --win10 --devices');
```
nikhilkh: does the cordova run --list command need an update? It will be nice to list both set of devices/emulators for Win10 & Win8.1 rather than have to specify an argument to pick and chose for listing purposes.

spec/unit/deployment.spec.js
robpaveza: Yes it is.  Jasmine 2 has some new features with respect to async, but yeah, this is pretty gross.

template/cordova/lib/deployment.js
robpaveza: It's "private" data - in other words, DeviceInfo promises an index, name, and type property.  The other properties that hang off of them are specific to the thing that created them.  Since we don't have Symbols yet, these will have to do.

template/cordova/lib/package.js
robpaveza: I don't understand this comment.  Code comments should specify why you do something when it isn't immediately clear, not just repeat the code in comment form.

template/cordova/lib/run.js
robpaveza: Because Windows 8.1 tools currently can't deploy to an 8.1 app to a Windows 10 device.  Hence the parenthetical and the following line.

template/cordova/lib/deployment.js
robpaveza: We use strings extensively throughout the codebase already.

template/cordova/lib/target-list.js
robpaveza: Yeah, this is a complicated question, and I'm not sure what the right answer is right now.

The Windows 8.1 tool actually enumerates Windows 10 emulators, but it can't deploy to them.  Conversely, the Windows 10 tool does not presently enumerate Windows 10 emulators (because it also can't deploy to them).  The Windows 8.1 tool is fast, the Windows 10 tool is slow (it does network probing).  We could possibly display both, but most likely we would end up simply having the following list, for example:

0. Device (device) [Windows 8.1 tools]
1. Mobile 4 inch 512MB 10.0.10176.0 (emulator) [Windows 8.1 tools]
2. Mobile 4 inch 1GB 10.0.10176.0 (emulator) [Windows 8.1 tools]
3. Mobile 5 inch 1GB 10.0.10176.0 (emulator) [Windows 8.1 tools]
4. Mobile 6 inch 2GB 10.0.10176.0 (emulator) [Windows 8.1 tools]
5. Windows Phone 8.1 4 inch 512MB (emulator) [Windows 8.1 tools]
6. Windows Phone 8.1 4 inch 1GB (emulator) [Windows 8.1 tools]
7. Windows Phone 8.1 5 inch 1GB (emulator) [Windows 8.1 tools]
8. Windows Phone 8.1 6 inch 2GB (emulator) [Windows 8.1 tools]
9. Lumia 1520 (RM-940) (device) [Windows 10 tools]

0 and 9 end up being the same thing, which is problematic.  I just don't know which is better: to repeat devices, or to have an option that picks "is my target device Windows 8.1 or Windows 10?"  I landed on the option choice, for compatibility reasons (opt into the new behavior).

template/cordova/lib/package.js
robpaveza: Yeah....  I didn't want to mess with this code and end up breaking something downstream.  It seems to have been working for us up until now.

template/cordova/lib/package.js
robpaveza: concat() returns a copy of the original array.  http://swingpants.com/2009/03/12/fastest-way-to-copy-an-array-concat-or-slice0/

template/cordova/lib/deployment.js
robpaveza: This function is based on spawn.js, modified only to capture stdout/stderr.  The comment is a fair question, but I thought I'd preserve consistency in this case.

template/cordova/lib/package.js
robpaveza: Heh, that's ActionScript, but since it all comes from ES3 it's still true.

template/cordova/lib/package.js
nikhilkh: It's not clear from looking at the code what the literal `true` stands for unless I got to definition of the function `installAppPackage`.

template/cordova/lib/run.js
nikhilkh: Fair enough - I remember we faced the same thing with xapDeploy and appDeploy. :)

template/cordova/lib/target-list.js
nikhilkh: Fair enough - this is more complicated than it should be. :8ball:

template/cordova/lib/deployment.js
nikhilkh: I remember in C# & C++ 'private' data was often represented with a single '_' (underscore) and not two underscores.

template/cordova/lib/package.js
TimBarham: :+1: 

It's a common recommendation for literal Boolean arguments in particular to name them at the call point to aid readability. Since JavaScript doesn't support named arguments, the usual approach is to include an inline comment, so:
```js
return deploymentTool.installAppPackage(package.appx, target,  /*shouldLaunch*/ true, /*somethingElse*/ false);
```

robpaveza: Ah, now I understand.  I have pushed a new change that will address this feedback.

nikhilkh: Thanks! :clap:

muratsu:
lgtm

PR #101, opened 2015-07-15
CB-9359: Adds support for .appxbundle creation

robpaveza:
.appxbundle files are convenient app packages which can be uploaded to the
Windows Store.  They contain debug or release (typically release) versions
of more than one of the x86, x64, and ARM architecture builds.

To enable bundling, simply build for multiple architectures:

    cordova build windows -- --archs="x86 x64 arm"

Bundling is disabled if `anycpu` is one of the architectures (or is the
default architecture).  You can explicitly disable bundling by passing the
`--no-bundle` flag:

    cordova build windows -- --archs="x86 x64 arm" --no-bundle

Bundling is not supported for Windows 8 as the feature is not available
for that platform in general.


robpaveza:
I see a different history on AppVeyor than on my machine, I'm going to close the PR and re-create.

PR #102, opened 2015-07-15
CB-9359: Adds support for .appxbundle creation

robpaveza:
.appxbundle files are convenient app packages which can be uploaded to the Windows Store.  They contain debug or release (typically release) versions of more than one of the x86, x64, and ARM architecture builds.

To enable bundling, simply build for multiple architectures:

    cordova build windows -- --archs="x86 x64 arm"

Bundling is disabled if `anycpu` is one of the architectures (or is the default architecture).  You can explicitly disable bundling by passing the `--no-bundle` flag:

    cordova build windows -- --archs="x86 x64 arm" --no-bundle

Bundling is not supported for Windows 8 as the feature is not available for that platform in general.

### DIFF
--- a/template/cordova/lib/MSBuildTools.js
+++ b/template/cordova/lib/MSBuildTools.js
@@ -29,7 +29,7 @@ function MSBuildTools (version, path) {
     this.path = path;
 }
 
-MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch) {
+MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch, otherConfigProperties) {
     console.log('Building project: ' + projFile);
     console.log('\tConfiguration : ' + buildType);
     console.log('\tPlatform      : ' + buildarch);
@@ -37,6 +37,13 @@ MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch) {
     var args = ['/clp:NoSummary;NoItemAndPropertyList;Verbosity=minimal', '/nologo',
     '/p:Configuration=' + buildType,
     '/p:Platform=' + buildarch];
+
+    if (otherConfigProperties) {
+        var keys = Object.keys(otherConfigProperties);
+        keys.forEach(function(key) {
+            args.push('/p:' + key + '=' + otherConfigProperties[key]);
+        });
+    }
 
     return spawn(path.join(this.path, 'msbuild'), [projFile].concat(args));
 };
@@ -50,6 +57,16 @@ module.exports.findAvailableVersion = function () {
         var msbuildTools = versions[0] || versions[1] || versions[2];
 
         return msbuildTools ? Q.resolve(msbuildTools) : Q.reject('MSBuild tools not found');
+    });
+};
+
+module.exports.findAllAvailableVersions = function () {
+    var versions = ['14.0', '12.0', '4.0'];
+
+    return Q.all(versions.map(checkMSBuildVersion)).then(function(unprocessedResults) {
+        return unprocessedResults.filter(function(item) {
+            return !!item;
+        });
     });
 };
 
--- a/template/cordova/lib/build.js
+++ b/template/cordova/lib/build.js
@@ -23,6 +23,7 @@ var Q     = require('Q'),
     shell = require('shelljs'),
     utils = require('./utils'),
     prepare = require('./prepare'),
+    package = require('./package'),
     MSBuildTools = require('./MSBuildTools'),
     ConfigParser = require('./ConfigParser'),
     fs = require('fs');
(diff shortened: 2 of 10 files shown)


template/cordova/lib/MSBuildTools.js
```diff
@@ -38,6 +38,13 @@ MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch) {
     '/p:Configuration=' + buildType,
     '/p:Platform=' + buildarch];
 
+    if (otherConfigProperties) {
```
nikhilkh: This approach is not consistent with other ways we've done this in the code here. For packaging, we've leaned on generating a targets file every time we build. We should use that same mechanism here: https://github.com/apache/cordova-windows/blob/master/template/cordova/lib/prepare.js#L74

Also, one of the downsides of this approach is that IDE <-> command line interop is broken. It's desirable for the IDE build to have the same results as the command line build. By adding an additional property at command line this can't be guaranteed.

robpaveza: I'll look into the .targets approach.

However, I don't agree with the IDE-CLI interop dichotomy.  In the IDE, you almost always are building a single arch+mode target pair at a time (e.g., x86-debug).  In VS, you can generate app package bundles, but you have to go through a particular menu to do so, not use the typical MSBuild-invoking "Build" step workflow.  The same works here as well.

The change was designed 100% to be just like the "Create App Packages" feature, just turned on by default when building arch-specific packages.

dblotsky:
Small question: do you think it's better to have an argument contain multiple values with an arbitrary separator (in other tools we sometimes use commas, or pipes, or also spaces), or is it better to have the argument be repeatable to likewise specify a list?

robpaveza:
I don't really have a preference. We could say be consistent with msbuild and do a pipe separator but I don't think that is terribly valuable. 

The archs argument is already existing and I don't think a redesign is in scope.

dblotsky:
Sounds fair!

template/cordova/lib/build.js
```diff
@@ -134,6 +141,12 @@ function parseAndValidateArgs(argv) {
         config.phone = args.phone ? true : false;
         config.win = args.win ? true : false;
         config.projVerOverride = args.appx;
+        // only set config.bundle if > 1 architecture none of which is any cpu
+        if (config.buildArchs.length > 1 && config.buildArchs.indexOf('anycpu') === -1 && config.buildArchs.indexOf('any cpu') === -1) {
```
muratsu: `config.buildArchs.length > 1` :+1:

template/cordova/lib/build.js
```diff
@@ -229,11 +268,66 @@ function buildTargets(msbuild, config) {
...
     }, Q());
+
+    if (shouldBundle) {
+        return buildsCompleted.then(function() {
+            // msbuild isn't capable of generating bundles unless you enable bundling for each individual arch
+            // However, that generates intermediate bundles, like "CordovaApp.Windows10_0.0.1.0_x64.appxbundle"
+            // We need to clear the intermediate bundles, or else "cordova run" will fail because of too
+            // many .appxbundle files.
+
+            console.log('Clearing intermediates...');
```
nikhilkh: Consider factoring this out - this method is getting gigantic.

template/cordova/lib/build.js
```diff
@@ -229,11 +268,66 @@ function buildTargets(msbuild, config) {
...
+            var appPackagesPath = path.join(ROOT, 'AppPackages');
+            var childDirectories = shell.ls(path.join(appPackagesPath, '*')).map(function(pathName) {
+                return { path: pathName, stats: fs.statSync(pathName) };
+            }).filter(function(fileInfo) {
+                return fileInfo.stats.isDirectory();
+            });
+
+            if (childDirectories.length === 0) {
+                throw new Error('Could not find a completed app package directory.');
```
muratsu: should we Q.reject instead?

template/cordova/lib/build.js
```diff
@@ -229,11 +268,66 @@ function buildTargets(msbuild, config) {
...
+            var archSearchString = bundleTerms.replace(/\|/g, '_') + (config.buildType === 'debug' ? '_debug' : '') + '.appxbundle';
+            var filesToDelete = shell.ls(path.join(outputDirectory.path, '*.appx*')).filter(function(appxbundle) {
+                var isMatch = appxbundle.indexOf(archSearchString) === -1;
+                if (!isMatch) {
+                    finalFile = appxbundle;
+                }
+                return isMatch;
+            });
+            filesToDelete.forEach(function(file) {
+                shell.rm(file);
```
muratsu: should we fail fast if we have trouble deleting a file?

template/cordova/lib/run.js
```diff
@@ -104,12 +104,16 @@ module.exports.run = function (argv) {
         projectType = 'windows80';
     }
 
+    if (projectType === 'windows80') {
+        // Don't enable bundling for Windows 8.
+        argv.push('--no-bundle');
```
muratsu: argv might already contain --no-bundle. this potentially adds an extra argument

nikhilkh:
This is awesome! LGTM.

dblotsky:
Ping on this. Please rebase atop master and verify that tests pass.

PR #103, opened 2015-07-15
add build matrix

purplecabbage:
Testing to see if this resolves our AppVeyor issues with recent builds.


purplecabbage:
Okay, we have the same issues that cordova-js was having with node v0.12 failing and v0.10 succeeding.
What is the proper way to fix this?

PR #105, opened 2015-07-24
CB-9408: Added a `windows-packageVersion` attribute to the `<widget>`

robpaveza:
...element which enables Windows to have a separate, independent binary
version than other platforms.  This attribute corresponds to similar
attributes `android-versionCode` and `ios-CFBundleVersion`.


omefire:
LGTM!

GuilhemBarboux:
Hi,

when i add windows platform on cordova project, the version number in .appxmanifests aren't update on identity tag. It's normaly ?

Thks

PR #106, opened 2015-07-25
CB-9410: Added preferences for Windows Store ingestion.

robpaveza:
`WindowsStoreDisplayName` and `WindowsStorePublisherName` are the new properties.  If they are not provided, we fall back to the defaults provided by `<widget>`.

### DIFF
--- a/spec/unit/Prepare.Win10.spec.js
+++ b/spec/unit/Prepare.Win10.spec.js
@@ -116,7 +116,19 @@ describe('Min/Max UAP versions are correctly read from the config file.', functi
   * Unit tests for validating default ms-appx-web:// URI scheme in Win10
   * (for the function applyCoreProperties) from prepare.js.
   **/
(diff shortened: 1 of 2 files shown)


nikhilkh:
cordova-lib does some replacement using the `displayname` and it might not work with this change. I'm not tally familiar with what cordova-lib is doing - but would be good to review.

nikhilkh:
@robpaveza - Did you get a chance to look at the comment I mentioned?

robpaveza:
It only goes into the case if it's an "old project template."  This change is only expected for newer Cordova projects.  (src/cordova/metadata/windows_parser.js line 73).  In other words, if you have an out-of-date cordova-windows, you don't get the updated behavior.

nikhilkh:
Ah makes sense - thanks. At what point can we remove this code from cordova-lib - not sure if support for older project template is that relevant now? We could display for people to update their `windows` platform project.

robpaveza:
I'm game.  Let's make a Jira for it.

PR #112, opened 2015-08-11
CB-9482: Modifies the deployment code to account for problems with

robpaveza:
emulators, which are currently blocking the CI from executing.  Modifying
the uninstall function to never fail appears to unblock the CI and allow
installation to emulators to continue unhindered.

### DIFF
--- a/template/cordova/lib/deployment.js
+++ b/template/cordova/lib/deployment.js
@@ -35,7 +35,7 @@ function run(cmd, args, opt_cwd) {
         child.stderr.on('data', function(s) { stderr += s; });
         child.on('exit', function(code) {
             if (code) {
-                d.reject(stderr);
+                d.reject({ message: stderr, code: code, toString: function() { return stderr; }});
             } else {
                 d.resolve(stdout);
             }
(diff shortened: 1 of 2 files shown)


dblotsky:
This seems fine in the interim, but let's make a note to keep the issue open until we can get a proper fix. By "proper", I mean things like checking for existence of a device before we try to uninstall, making uninstall more reliable, etc.

dblotsky:
I rebased and forgot to add "This closes #112" to 050910c, sorry. You may close the PR.

robpaveza:
There isn't a way to get a "proper" fix.  The tool doesn't have a mechanism to query for the presence of a device (AppDeployCmd /EnumerateDevices always lists "0. Device"), nor a way to query for the presence of a particular app.  The only thing we CAN do is try to uninstall the app, to ensure that it always gets uninstalled if it's there, and then to actually install the app.  (AppDeployCmd does have an "update" option, but it fails if the app isn't already installed).

The problem is with the SDK tool, and given that Windows Phone 8.1 is now a revision out of date, it's unlikely we'll get a fix.

PR #113, opened 2015-08-12
CB-9482: Upon further investigation of CI failures, we saw that

robpaveza:
AppDeployCmd also failed to install an app when:
 - The emulator had been off previously
 - It had been started with the AppDeployCmd /uninstall command
 - The next command was to then install an app

This change detects the condition (only for emulator targets) and retries
once.  This appears to correct the issue.

### DIFF
--- a/template/cordova/lib/deployment.js
+++ b/template/cordova/lib/deployment.js
@@ -24,6 +24,8 @@ var Q     = require('q'),
     /* jshint ignore:end */
     proc  = require('child_process');
 
+var E_INVALIDARG = 2147942487;
+
 // neither 'exec' nor 'spawn' was sufficient because we need to pass arguments via spawn
 // but also need to be able to capture stdout / stderr
 function run(cmd, args, opt_cwd) {
@@ -35,7 +37,7 @@ function run(cmd, args, opt_cwd) {
         child.stderr.on('data', function(s) { stderr += s; });
         child.on('exit', function(code) {
             if (code) {
-                d.reject(stderr);
+                d.reject({ message: stderr, code: code, toString: function() { return stderr; }});
             } else {
                 d.resolve(stdout);
             }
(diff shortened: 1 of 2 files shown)


template/cordova/lib/deployment.js
```diff
@@ -194,13 +194,28 @@ AppDeployCmdTool.prototype.enumerateDevices = function() {
...
         command += 'launch';
     }
 
-    return run(this.path, [command, pathToAppxPackage, '/targetdevice:' + targetDevice.__shorthand]);
+    var that = this;
+    var result = run(this.path, [command, pathToAppxPackage, '/targetdevice:' + targetDevice.__shorthand]);
+    if (targetDevice.type === 'emulator' && arguments.length < 6) {
+        result = result.then(null, function(e) {
+            // CB-9482: AppDeployCmd also reports E_INVALIDARG during this process.  If so, try to repeat.
+            if (e.code === 2147942487) {
```
dblotsky: Nitpick: Please define this as a constant.

template/cordova/lib/deployment.js
```diff
@@ -194,13 +194,28 @@ AppDeployCmdTool.prototype.enumerateDevices = function() {
...
 AppDeployCmdTool.prototype.installAppPackage = function(pathToAppxPackage, targetDevice, shouldLaunch, shouldUpdate, pin) {
     var command = shouldUpdate ? '/update' : '/install';
     if (shouldLaunch) {
         command += 'launch';
     }
 
-    return run(this.path, [command, pathToAppxPackage, '/targetdevice:' + targetDevice.__shorthand]);
+    var that = this;
+    var result = run(this.path, [command, pathToAppxPackage, '/targetdevice:' + targetDevice.__shorthand]);
+    if (targetDevice.type === 'emulator' && arguments.length < 6) {
```
dblotsky: What's special about `arguments.length` and 6?

template/cordova/lib/deployment.js
```diff
@@ -35,7 +35,7 @@ function run(cmd, args, opt_cwd) {
         child.stderr.on('data', function(s) { stderr += s; });
         child.on('exit', function(code) {
             if (code) {
-                d.reject(stderr);
+                d.reject({ message: stderr, code: code, toString: function() { return stderr; }});
```
dblotsky: Is this a better way to log, or was it only used in debugging?

template/cordova/lib/deployment.js
robpaveza: OK, I'll do so in the updated PR, but I don't see that `var E_INVALIDARG = ...; if (e.code === E_INVALIDARG)` is necessarily better.

template/cordova/lib/deployment.js
robpaveza: This isn't logging.  This is providing an object instead of a string to the failure case of a promise.

template/cordova/lib/deployment.js
robpaveza: It's called out in the comment immediately preceding the function definition.  There are 5 formal arguments specified to the function, so when the function is called normally, arguments.length = 5.  In that case, we compose an error handler onto result; the composed error handler is only called if AppDeployCmd fails.  That handler checks the code to see if it's `E_INVALIDARG`, and if so, re-calls the function, this time passing an additional parameter.

robpaveza: In the update, I'll just re-call run().

template/cordova/lib/deployment.js
dblotsky: Whoops, I totally read `d.reject()` as `console.log`.

PR #114, opened 2015-08-13
CB-8936 Windows logs: Improvements

alsorokin:
https://issues.apache.org/jira/browse/CB-8936

- Now also showing WWATerminateApplication
- Script now exits if no channels can be enabled
- Log cleanup (hide line and column fields if they are both 0)
- Slight code refactoring

### DIFF
--- a/template/cordova/lib/log.js
+++ b/template/cordova/lib/log.js
@@ -24,22 +24,40 @@ var path    = require('path'),
     shelljs = require('shelljs'),
     ROOT    = path.join(__dirname, '..', '..');
 
-var appTracingLogInitialState = null,
-    adminLogInitialState = null;
+var appTracingInitialState = null,
+    appTracingCurrentState = null,
+    adminInitialState = null,
+    adminCurrentState = null;
 
 /*
  * Gets windows AppHost/ApplicationTracing and AppHost/Admin logs
  * and prints them to console
  */
 module.exports.run = function() {
-    getLogState('Microsoft-Windows-AppHost/ApplicationTracing').then(function (state) {
-        appTracingLogInitialState = state;
-        return getLogState('Microsoft-Windows-AppHost/Admin');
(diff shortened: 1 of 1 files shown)


nikhilkh:
LGTM

template/cordova/lib/log.js
```diff
@@ -114,32 +132,47 @@ function parseEvents(output) {
...
-            message: getElementValue(event, './UserData/WWADevToolBarLog/Message'),
-            displayName: getElementValue(event, './UserData/WWAUnhandledApplicationException/DisplayName'),
-            appName: getElementValue(event, './UserData/WWAUnhandledApplicationException/ApplicationName'),
-            errorType: getElementValue(event, './UserData/WWAUnhandledApplicationException/ErrorType'),
-            errorDescription: getElementValue(event, './UserData/WWAUnhandledApplicationException/ErrorDescription'),
-            stackTrace: getElementValue(event, './UserData/WWAUnhandledApplicationException/StackTrace'),
+            channel:          getElementValue(event, './System/Channel'),
+            timeCreated:      getElementValue(event, './System/TimeCreated', 'SystemTime'),
+            pid:              getElementValue(event, './System/Execution', 'ProcessID'),
+            proc:             getElementValue(event, './UserData/WWADevToolBarLog/DisplayName'),
```
nikhilkh: Nit: instead of proc & 'Process' - we should call it 'App Name'

nikhilkh:
These logs are global and hence, we should filter to the application ID for the current app. I believe we should already have logic to parse the application ID.

nikhilkh:
When I press Ctrl+C I ran into an issue - the node process got a busy hang with pegged CPU and the last log line was - 'Disabling channel Microsoft-Windows-AppHost/ApplicationTracing'. Is the synchronous shelljs.exec causing problem when the process is terminating?

PR #116, opened 2015-08-17
CB-9499: There is a run failure when trying to deploy an x64 app when…

robpaveza:
… running on an x86 version of Node.  This change addresses the problem by modifying the run process by detecting an x64 app package as the primary, and setting the environment architecture to x64 before invoking PowerShell.

Also, small build item to point to the correct location of WinJS.

### DIFF
--- a/template/cordova/lib/package.js
+++ b/template/cordova/lib/package.js
@@ -215,15 +215,25 @@ module.exports.deployToDesktop = function (package, deployTarget) {
 
     return utils.getAppStoreUtils().then(function(appStoreUtils) {
         return getPackageName(path.join(__dirname, '..', '..')).then(function(pkgname) {
+
+            var oldArch;
             // uninstalls previous application instance (if exists)
             console.log('Attempt to uninstall previous application version...');
             return spawn('powershell', ['-ExecutionPolicy', 'RemoteSigned', 'Import-Module "' + appStoreUtils + '"; Uninstall-App ' + pkgname])
             .then(function() {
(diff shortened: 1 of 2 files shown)


purplecabbage:
lgtm!

template/cordova/lib/package.js
```diff
@@ -215,13 +215,20 @@ module.exports.deployToDesktop = function (package, deployTarget) {
...
             // uninstalls previous application instance (if exists)
             console.log('Attempt to uninstall previous application version...');
             return spawn('powershell', ['-ExecutionPolicy', 'RemoteSigned', 'Import-Module "' + appStoreUtils + '"; Uninstall-App ' + pkgname])
             .then(function() {
                 console.log('Attempt to install application...');
+                oldArch = process.env.PROCESSOR_ARCHITECTURE;
+                if (package.arch === 'x64') {
+                    process.env.PROCESSOR_ARCHITECTURE = 'AMD64';
+                }
                 return spawn('powershell', ['-ExecutionPolicy', 'RemoteSigned', 'Import-Module "' + appStoreUtils + '"; Install-App', utils.quote(package.script)]);
```
muratsu: If we throw here the PROCESSOR_ARCHITECTURE will remain changed. While this currently has no side effects, we probably want to have a catch promise and set the PROCESSOR_ARCHITECTURE to the original value since it's a shared value. LGTM otherwise.

robpaveza: Addressed by update.  Thx.

PR #119, opened 2015-09-02
CB-8936 Introduced --dump arg to log script

alsorokin:
https://issues.apache.org/jira/browse/CB-8936

### DIFF
--- a/template/cordova/lib/log.js
+++ b/template/cordova/lib/log.js
@@ -22,41 +22,64 @@ var path         = require('path'),
     et           = require('elementtree'),
     Q            = require('q'),
     cp           = require('child_process'),
-    ConfigParser = require('./ConfigParser.js');
+    ConfigParser = require('./ConfigParser.js'),
+    nopt         = require('nopt');
 
 // paths
 var platformRoot = path.join(__dirname, '..', '..'),
     projectRoot  = path.join(platformRoot, '..', '..'),
     configPath   = path.join(projectRoot, 'config.xml');
 
+//constants
+var APP_TRACING_LOG = 'Microsoft-Windows-AppHost/ApplicationTracing',
+    ADMIN_LOG       = 'Microsoft-Windows-AppHost/Admin';
+
 // variables
 var appTracingInitialState = null,
     appTracingCurrentState = null,
-    adminInitialState = null,
-    adminCurrentState = null,
+    adminInitialState      = null,
+    adminCurrentState      = null,
+    timers                 = [],
     appName;
 
 /*
  * Gets windows AppHost/ApplicationTracing and AppHost/Admin logs
  * and prints them to console
  */
-module.exports.run = function() {
-    getLogState('Microsoft-Windows-AppHost/Admin').then(function (state) {
+module.exports.run = function(args) {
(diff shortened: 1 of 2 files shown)


template/cordova/lib/log.js
```diff
@@ -30,33 +30,87 @@ var platformRoot = path.join(__dirname, '..', '..'),
...
  */
-module.exports.run = function() {
-    getLogState('Microsoft-Windows-AppHost/Admin').then(function (state) {
+module.exports.run = function(args) {
+    var dump = false, 
+        startTime = new Date(new Date().getTime() - 10 * 60 * 1000).toISOString(); // show last 10 minutes by default
+
+    if (args.length > 0) {
+        for (var i = 0; i < args.length; i++) {
+            switch (args[i]) {
```
nikhilkh: Ideally, we should use a library for this - the one that is used by cordova-windows should be re-used here instead of doing this on your own.

template/cordova/lib/log.js
```diff
@@ -30,33 +30,87 @@ var platformRoot = path.join(__dirname, '..', '..'),
...
+                case '--startTime':
```
nikhilkh: Instead of start time which is often hard to specify you should have last 'n' minutes or seconds: `--mins`

template/cordova/lib/log.js
```diff
@@ -30,33 +30,87 @@ var platformRoot = path.join(__dirname, '..', '..'),
...
+        var appTracingEvents, adminEvents;
+        getEvents(appTracingLog, startTime).then(function (evts) {
+            appTracingEvents = evts;
+            return getEvents(adminLog, startTime);
+        }).then(function(evts) {
+            adminEvents = evts;
+            appTracingEvents.concat(adminEvents)
+            .sort(function(evt1, evt2) {
+                if (evt1.timeCreated < evt2.timeCreated) {
+                    return -1;
```
vladimir-kotikov: if `evt*.timeCreated` is a Number, you might want to simplify this to `return evt1.timeCreated - evt2.timeCreated;`

alsorokin: it's actually a string, a time in ISO format, like that: 2011-10-05T14:48:00.000Z

template/cordova/lib/log.js
```diff
@@ -30,33 +30,87 @@ var platformRoot = path.join(__dirname, '..', '..'),
...
+                case '-h':
+                    module.exports.help();
+                    return;
+                default:
+                    console.warn('Ignoring unknown arg: ' + args[i]);
+            }
+        }
+    }
+
+    if (dump) {
```
vladimir-kotikov: The code for 'dump' case might look better if factored to separate method. The `run` method already looks too crowded.

template/cordova/lib/log.js
vladimir-kotikov: Oh, i see...

template/cordova/lib/log.js
purplecabbage: -d has been used in the past to mean debug, which gives more verbose output. 
Shortening 4:1 is not a big deal imo.

alsorokin:
Updated.

vladimir-kotikov:
LGTM

PR #120, opened 2015-09-07
Add support for back button on win10

EionRobb:
Win10 has support for backbutton through the SystemNavigationManager API 
https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.core.systemnavigationmanager.aspx 

Unfortunately there's some OS-level crashes that happen when running the app if it has been compiled with VS2012/2013, but they get detected with the userAgent check that the OS fakes when it runs the app container, so this code only works when compiling on VS2015 (or newer, I guess)

### DIFF
--- a/cordova-js-src/platform.js
+++ b/cordova-js-src/platform.js
@@ -38,6 +38,24 @@ module.exports = {
             channel.onActivated = cordova.addDocumentEventHandler('activated');
         }
         channel.onNativeReady.fire();
+        
+        // Only load this code if we're running on Win10 in a non-emulated app frame, otherwise crash \o/
+        if (parseInt(window.clientInformation.userAgent.match(/Windows NT ([0-9.]+)/)[1]) >= 10) {
+            var sysnavman = Windows.UI.Core.SystemNavigationManager.getForCurrentView();
+            // Inject a listener for the backbutton on the document.
+            var backButtonChannel = cordova.addDocumentEventHandler('backbutton');
+            backButtonChannel.onHasSubscribersChange = function() {
+                // If we just attached the first handler or detached the last handler,
+                // let native know we need to override the back button.
+                sysnavman.appViewBackButtonVisibility = (this.numHandlers == 1);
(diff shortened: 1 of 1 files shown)


sgrebnov:
I plan to test and merge it. @EionRobb  is there associated Jira issue available? Also, is there a special reason we use user agent instead of feature detection (`Windows.UI.Core.SystemNavigationManager.getForCurrentView`)?

EionRobb:
Feature detection is broken. It causes the ieframe.dll to crash outside of js and is uncatchable.

EionRobb:
To clarify, the method defines itself as existing, but any attempts to use it will crash the container. 

There isn't a Jira issue for it, that would be unnecessary duplication.

sgrebnov:
@EionRobb could you please verify the following change: 
```
parseInt(window.clientInformation.userAgent.match(/Windows NT ([0-9.]+)/)[1]) >= 10
->
navigator.appVersion.indexOf('MSAppHost/3.0') !== -1
```
to support Mobile 10 as well

https://github.com/apache/cordova-windows/commit/02a1ded1edb70d38f9259f8ab4543887514a19f8#diff-39f17a8caa1889e86840a9f4c1dd99b2R127

sgrebnov:
PS. changing `appViewBackButtonVisibility ` on Mobile 10 does not take any effect, standard phone back button works as expected (correctly handled using `backrequested` event)

EionRobb:
I wouldn't personally make it look for the "MSAppHost/3.0" string, maybe doing a regex match on that to see if it's version 3.0 or higher might be better?
```
parseInt(navigator.appVersion.match(/MSAppHost\/([0-9]+)/)[1]) >= 3
```
or something, just to futureproof it

sgrebnov:
Agree and thx! - I made it this way to be consistent w/ other place where we do similar thing
https://github.com/apache/cordova-windows/blob/master/cordova-js-src/platform.js#L74

PR #123, opened 2015-09-22
CB-9565: Build failure for Windows 10.  This error is a result of using

robpaveza:
This error is a result of using x64 Node and an incompatibility with x64 msbuild; the .NET Native compiler is x86 only (for where it can run).  As a result, when msbuild-x64 tries to use it, it fails.  This works around the issue by running x86 instead.

### DIFF
--- a/template/cordova/lib/MSBuildTools.js
+++ b/template/cordova/lib/MSBuildTools.js
@@ -77,7 +77,13 @@ function checkMSBuildVersion(version) {
         // fetch msbuild path from 'reg' output
         var path = /MSBuildToolsPath\s+REG_SZ\s+(.*)/i.exec(output);
         if (path) {
-            deferred.resolve(new MSBuildTools(version, path[1]));
+            path = path[1];
+            // CB-9565: Windows 10 invokes .NET Native compiler, which only runs on x86 arch,
+            // so if we're running an x64 Node, make sure to use x86 tools.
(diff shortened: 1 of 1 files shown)


nikhilkh:
@robpaveza  Can you please fix the failing test and merge this soon?

nikhilkh:
@robpaveza Any update on this?

nikhilkh:
@rakatyal  Can you please take a look at this one?

SomaticIT:
@robpaveza @nikhilkh @rakatyal 
Any news on this, it's a really painful error.

The error in tests is an issue in @robpaveza fix [here](https://github.com/MSOpenTech/cordova-windows/blob/8ea8a719848e296e2cd54b2c5479160a9467046c/template/cordova/lib/MSBuildTools.js#L83).

````javascript
if (version === '14.0' && path.indexOf('amd64')) {
    path = require('path').join(path, '..');
}
````
should be:
````javascript
if (version === '14.0' && path.indexOf('amd64') !== -1) {
    path = require('path').join(path, '..');
}
````
Could this be fixed ?

ghenry22:
Currently blocked preparing my first windows store submission by this one.  I use an external path to x64 node in visual studio 2015 so that it uses an updated version that doesn't freak out over path depth but then you hit this error when building for store submission fails.

ghenry22:
tested with the mod as in the comments, resolved the error for me

PR #127, opened 2015-10-23
remove publisherId since its overridden by the certficate during build

pke:
1. The publisherId set here is __always__ overridden by the `CN` from the projects associated certificate. That is by default, the Cordova certificate that has `CN=The Cordova Team`. Therefor the resulting `AppxManifest.xml` will always have `Publisher="CN=The Cordova Team"` set regardless of what one puts into `publisherId`.


pke:
There seem to be more places where this needs to be removed. Will create a more complete PR.

PR #128, opened 2015-10-23
Replace the "CordovaApp" description with the real description from config.xml

pke:
The description was read from the `config.xml` but never used to modify the apps manifest files.


pke:
Hmm something wrong with that. Have to test locally.

PR #132, opened 2015-11-09
CB-9828 Implements PlatformApi contract for WIndows platform

vladimir-kotikov:
Implementation for https://issues.apache.org/jira/browse/CB-9827

Main highlights:

* All logic from cordova-lib/windows_parser and plugman/windows_handler now moved into platform
* Platform is now able to do a prepare and plugin installation (through API)
* Console logging was replaced with `events` usage (shared w/ cordova) so log level now appied to platform as well as to the lib.

The updated platform compatible with old cordova versions (as it still exposes main functionality through batch/shell scripts).

Also the mobliespec application builds sucessfully and mobilespec tests are passing without any regressions.


bin/lib/check_reqs.js
```diff
@@ -209,7 +209,7 @@ function getInstalledPhoneSdks () {
         }
     })
     .then(function () {
-        return exec('reg query "HKLM\\SOFTWARE\\Microsoft\\Microsoft SDKs\\Windows\\v10.0" /v InstallationFolder /reg:32');
+        return spawn('reg','query','HKLM\\SOFTWARE\\Microsoft\\Microsoft SDKs\\Windows\\v10.0','/v','InstallationFolder','/reg:32');
```
jasongin: Why aren't the arguments passed as an array here as above?

jasongin:
I looked through all of the PR and didn't find anything concerning other than the one thing I commented on. But it's a lot of code in an area I'm not at all familiar with.

TimBarham:
Generally looks fine to me apart from some misplaced "Android" references I mention inline. Also, it concerns me that `ConsoleLogger` is duplicated across each platform - isn't this something that can live in `cordova-common`? Why aren't we using the same functionality as `cordova-cli`?

vladimir-kotikov:
Addressed notes and fixed compatibility w/Node@0.12

Regarding logging, agree, we should move it common, but i propose to follow more general way - add necessary functionality to 'common', release it and update platforms (this affects iOS and Android as well) to use released 'common'. @TimBarham, @jasongin, thoughts?

jasongin:
Refactoring logging into common sounds good, but we shouldn't block this change on it.

vladimir-kotikov:
This was merged in [58047a3](https://git1-us-west.apache.org/repos/asf?p=cordova-windows.git;a=commit;h=58047a3daf21dbd046a6354bf748b83f3ae7b5b9)

PR #133, opened 2016-01-04
CB-10292: Windows platform support for next version of VS/MSBuild

jasongin:
Reference the JIRA issue for more explanation.

The targets filter for 15.0 is currently the same as 14.0, but I copied to a separate function for clarity and in case it ever changes while 15.0 is being developed further.

### DIFF
--- a/template/cordova/lib/MSBuildTools.js
+++ b/template/cordova/lib/MSBuildTools.js
@@ -50,18 +50,18 @@ MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch, o
 
 // returns full path to msbuild tools required to build the project and tools version
 module.exports.findAvailableVersion = function () {
-    var versions = ['14.0', '12.0', '4.0'];
+    var versions = ['15.0', '14.0', '12.0', '4.0'];
 
     return Q.all(versions.map(checkMSBuildVersion)).then(function (versions) {
         // select first msbuild version available, and resolve promise with it
-        var msbuildTools = versions[0] || versions[1] || versions[2];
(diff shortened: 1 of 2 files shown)


jasongin:
@vladimir-kotikov please review, and commit if it looks good.

template/cordova/lib/MSBuildTools.js
```diff
@@ -50,11 +50,11 @@ MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch, o
 
 // returns full path to msbuild tools required to build the project and tools version
 module.exports.findAvailableVersion = function () {
```
vladimir-kotikov: We would also need to update [`findAllAvailableVersions` method](https://github.com/apache/cordova-windows/pull/133/files#diff-cb73739fe4756cb409aeb5941bbb8e47R64)

vladimir-kotikov:
Looks like [`getMsBuildForTargets`](https://github.com/jasongin/cordova-windows/blob/CB-10292/template/cordova/lib/build.js#L506) also requires update, otherwise we won't use 15.0 for build even if it is installed.

template/cordova/lib/MSBuildTools.js
jasongin: Done.

jasongin:
I updated a couple more places to support 15.0 in addition to 14.0. While those changes weren't necessary to build a cordova project from VS in the scenario we tested, it's more consistent and will probably handle other scenarios more correctly.

For now, the next version of VS still supports all the same targets as 2015. If that ever changes (e.g. it drops support for targeting Windows 8.1) then we'll need to update the logic.

PR #134, opened 2016-01-05
CB-10193 Add BOM to www files at build stage instead of prepare

vladimir-kotikov:
This PR moves adding BOM to JS files from prepare to build stage. This should fix the potential problem (as explained in [mailing list discussion](http://markmail.org/thread/cg5s2jvxvyyqcqpd)) when www files already contain BOM signatures at the time when 'after_prepare' hook ([or simulated 'pre_package' hook](https://github.com/apache/cordova-lib/commit/4478eaf0783a2f73962e5911a0c26cc4db1da241)) is fired.

JIRA issue: [CB-10193](https://issues.apache.org/jira/browse/CB-10193)


sgrebnov:
lgtm

PR #136, opened 2016-01-13
CB-10344 Fixed icons and splashscreens parsing

alsorokin:
https://issues.apache.org/jira/browse/CB-10344


sgrebnov:
looks great, thank you a lot - will merge after fixing core windows install issue.

PS. Can you also take a look if we can auto test this case please.

PR #138, opened 2016-01-14
CB-10234 Better error message when Windows10 requires 'arch' flag

sgrebnov:
This patch adds the following error message:
image

Original message is showed below:
image


nikhilkh:
@rakatyal FYI - LGTM!

rakatyal:
Looks good. Thanks Sergey.

PR #139, opened 2016-01-18
don't copy resource-files at plugin install

biasmv:
Instead, reference the file in the plugin directory. This allows to have multiple
source files map to the same destination file inside the build directory and
choose the appropriate file based on the build configuration, e.g. architecture.

This implements a fix for CB-10326

### DIFF
--- a/template/cordova/lib/JsprojManager.js
+++ b/template/cordova/lib/JsprojManager.js
@@ -108,19 +108,19 @@ jsprojManager.prototype = {
         });
     },
 
-    addResourceFileToProject: function (relPath, targetConditions) {
-        events.emit('verbose', 'jsprojManager.addResourceFile(relPath: ' + relPath + ', targetConditions: ' + JSON.stringify(targetConditions) + ')');
+    addResourceFileToProject: function (sourcePath, destPath, targetConditions) {
+        events.emit('verbose', 'jsprojManager.addResourceFile(sourcePath: ' + sourcePath + ', destPath: ' + destPath + ', targetConditions: ' + JSON.stringify(targetConditions) + ')');
 
         // add hint path with full path
         var link = new et.Element('Link');
-        link.text = relPath;
+        link.text = destPath;
         var children = [link];
 
         var copyToOutputDirectory = new et.Element('CopyToOutputDirectory');
         copyToOutputDirectory.text = 'Always';
         children.push(copyToOutputDirectory);
 
-        var item = createItemGroupElement('ItemGroup/Content', relPath, targetConditions, children);
+        var item = createItemGroupElement('ItemGroup/Content', sourcePath, targetConditions, children);
         this._getMatchingProjects(targetConditions).forEach(function (project) {
             project.appendToRoot(item);
         });
@@ -128,7 +128,6 @@ jsprojManager.prototype = {
 
     removeResourceFileFromProject: function (relPath, targetConditions) {
         events.emit('verbose', 'jsprojManager.removeResourceFile(relPath: ' + relPath + ', targetConditions: ' + JSON.stringify(targetConditions) + ')');
-
         this._getMatchingProjects(targetConditions).forEach(function (project) {
             project.removeItemGroupElement('ItemGroup/Content', relPath, targetConditions);
         });
--- a/template/cordova/lib/PluginHandler.js
+++ b/template/cordova/lib/PluginHandler.js
@@ -25,6 +25,13 @@ var shell = require('shelljs');
 var events = require('cordova-common').events;
 var CordovaError = require('cordova-common').CordovaError;
 
+// returns relative file path for a file in the plugin's folder that can be referenced 
+// from a project file. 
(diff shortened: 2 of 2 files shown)


nikhilkh:
I don't completely understand why you are packaging '.dll' as a resource file? Can you explain your scenario?

biasmv:
Yeah, that was confusing to me as well. My plugin uses an unmanaged native dll that I'm referencing from C# code. As described in the links below, it's not possible to reference these unmanaged dlls from the VS project the same way that COM libs or managed libraries can be referenced. Instead one must manually copy the files to the output directory. That's exactly what resource-file is here for, so it would fit my use-case perfectly.

http://http://stackoverflow.com/questions/5107694/
https://social.msdn.microsoft.com/Forums/en-US/f7824f2d-0d4b-4bf0-8341-fd0799e6ab1c/whats-the-best-way-to-add-unmanaged-dlls-to-a-project?forum=csharpide

Best,
Marco

sgrebnov:
I like the idea, but I think we will miss `target` attribute support as resource-file will be always copied to root (pls correct me if I'm wrong). In some cases you will want to use `resource-file` to specify some asset which is available for the app in runtime (image/json/xml/etc).

https://cordova.apache.org/docs/en/dev/plugin_ref/spec.html#link-resource-file-and-header-file-elements

Also, I believe using target attribute you will be able to get it working, for example 

```
<resource-file src="x86/foo.dll" target="x86/foo.dll" arch="x86" />
<resource-file src="arm/foo.dll" target="arm/foo.dll" arch="arm" />
<resource-file src="x64/foo.dll" target="x64/foo.dll" arch="x64" />
```

sgrebnov:
@nikhilkh - looks like the doc describing `target` support on windows and corresponding example exist in dev version only.We should make it released.

biasmv:
@sgrebnov: I've verified that it's still possible to have files in non-root folder locations. I've inserted the following into my project (which is essentially what would be generated by resource-file with the patch):

```
<Content Include="file.txt">
   <Link>somesubdirectory/file.txt</Link>
   <CopyToOutputDirectory>Always</CopyToOutputDirectory>
</Content>
```

The file is correctly placed in somesubdirectory and not the output directory root folder. So the patch does not prevent the resource element to be used that way.

The suggested use of the target attribute would still mean that the files land in different locations depending on the architecture. I'd need to change the C# library to somehow load the unmanaged dll from from a different location, depending on what device it's running on. I'd prefer if I could just copy the file with the correct architecture to the output directory without having to do any runtime checks.

Also, since I'm referencing the file from the appxmanifest as part of an extension (see the original ticket https://issues.apache.org/jira/browse/CB-10326), I'd need to change the path in the appxmanifest depending on the architecture currently being built.

sgrebnov:
Hi @biasmv - I'll take a look and respond today or tomorrow, thank you for clarification

biasmv:
Cool, thanks!

sgrebnov:
I've tried this approach and it seems to work very well. The only two things I've noticed are:
1. Generated `Include` is absolute, this could be a problem for remote build tools and when you move the app or send someone else. But I think this could be easily fixed.
```
<Content Condition="'$(Platform)'=='x64'" Include="C:\temp\resCopyTest\plugins\cordova-plugin-dialogs\dll\x64\foo.dll">
    <Link>x64/foo.dll</Link>
    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
</Content>
```
2. Includes to files which are outside the VS project are showed like linked folder is missing so this may confuse people
image

But in general this works well and should simplify working with arch-specific native dependencies a lot. File is correctly placed in generated package so there should no be any runtime issues

sgrebnov:
Here is my +1, but it will be great if someone else can take a look on this change as well.

TimBarham:
I've taken a look too, and I agree this seems the best approach. Changes look good to me, except I agree with your comment @sgrebnov that we should use a relative path rather than absolute.

Regarding the red "x" on the folder name: this seems to be a limitation of a `jsproj` (that they don't fully support links, at least in the VS UI). If you create an equivalent link in a `csproj`, you don't get the red "x". If it indeed builds fine, then we can live with that.

However, there is one way we can avoid the red "x": actually create any folder defined in the `target` path (if there is one). It would be empty, but would make for less confusion in the VS UI.

biasmv:
Thanks for taking a look! I agree that it would be better to use relative paths. At the moment, csproj files are also referenced using absolute paths, so one would have to convert them to relative paths, too before being able to move around the generated projects.

biasmv:
Hi,

just to follow up on this. I've changed the code to use relative paths. We have been using this version for a while for testing our plugin and it has been working really well. 

Is there anything to be done before this can be merged?

Best,
Marco

sgrebnov:
I'll test new version and merge later today

sgrebnov:
Tested and merged, @biasmv  Marco - thx for the patch!

biasmv:
That's fantastic! Thanks for merging.

PR #141, opened 2016-01-20
CB-8056 Implement splashscreen for Windows platform

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-8056)

Moved the implementation to the platform as per plugins triage recommendation and due to the need to hook into `activated` event for delaying the initial splashimage (`e.setPromise`).


purplecabbage:
Is there a link to the conversation around this change? ( I assume it is [1] )
Ultimately there are 2 pieces to splash-screen : 
1. the platform/os splash screen built into every app
2. the cordova specific plugin which processes preferences

My understanding is that if you do NOT have the plugin installed, then the platform/os splash is presented and dismissed automatically once the app has launched. The image(s) used for this can and should be controlled by preferences in config.xml and this should be done at build time, moving files, renaming files, modifying app manifests, ...

If the splash-screen plugin is installed, we gain the ability to delay dismissing the splash screen ( by emulating it with a full screen image which quickly replaces the platform/os version )

This is the way things are currently, and I DO completely support moving splash-screen management back into the platforms so it can be processed before plugins are loaded, I just want to make sure that we are moving on this for all platforms and that there has been enough discussion/agreement around it.
I think I expected that each of the plugin audit proposals should spawn it's own detailed proposal, but I may be over administering things.  As long as we all agree that this is the way forward, and this will be done for all platforms; +1

[1] https://github.com/cordova/cordova-discuss/pull/28/files#diff-e1c3c74bda464d02680ae112f2927180R175

daserge:
@purplecabbage, thanks for reviewing!

The main reason to move the code to the platform was the need to hook into activated event so that the initial splash image could be delayed until the plugin is configured.

I'm not sure when/if other platforms are going to be rewritten in this way.

SunboX:
Two comments on this:

* calling `navigator.splashscreen.hide();` does not hide the splash screen (Windows 10, Lumia 950, build Cordova 6.0 as UWP 10)
* calling `StatusBar.show();` should not show the status bar while the splash screen is shown (and should show the status bar automatically after the splash screen gets hidden)

SunboX:
> * calling navigator.splashscreen.hide(); does not hide the splash screen (Windows 10, Lumia 950, build Cordova 6.0 as UWP 10)

Ok, this was my fault. I had the splashscreen plugin installed additionally. Removed it, now it works as expected. Thanks for your work!

purplecabbage:
This now has conflicts, can we get a rebase and a merge? lgtm!

daserge:
@SunboX, @chrigi, I've rebased and updated the PRs, could you please try how it works for you?
`cordova platform add https://github.com/MSOpenTech/cordova-windows#d6988f5c863ec3ba5cde33a678a4b93badc64395`
`cordova plugin add https://github.com/MSOpenTech/cordova-plugin-splashscreen#acb45f3a386bbb80d1d52cbe9d26f3de2617bfe0`

Sample config.xml:
```
    <preference name="SplashScreenBackgroundColor" value="#292826" />
    <preference name="SplashScreenSpinnerColor" value="#d03932" />

    <preference name="FadeSplashScreenDuration" value="500" />

    <preference name="AutoHideSplashScreen" value="true" />

    <preference name="FadeSplashScreen" value="true" />

    <preference name="SplashScreenDelay" value="2500" />

    <platform name="windows">
        <!-- images are determined by width and height. The following are supported -->
        <splash src="res/screen/windows/splashscreen.png" width="620" height="300"/>
        <splash src="res/screen/windows/splashscreenphone.png" width="1152" height="1920"/>
    </platform>
```

daserge:
@chrigi, note that I haven't addressed this:

> It does not set the statusbar color to the splashScreenBackgroundColor

I will file a separate issue for this when the PRs are merged.

purplecabbage:
Verified, and merged. Thanks!

PR #142, opened 2016-01-26
CB-10440 Add CSS color names support for BackgroundColor on Windows

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10440)

### DIFF
--- a/template/cordova/lib/AppxManifest.js
+++ b/template/cordova/lib/AppxManifest.js
@@ -382,7 +382,15 @@ AppxManifest.prototype.getCapabilities = function () {
     });
 };
 
+function isCSSColorName(color) {
+    return color.indexOf('0x') === -1 && color.indexOf('#') === -1;
+}
+
 function refineColor(color) {
+    if (isCSSColorName(color)) {
+        return color;
+    }
+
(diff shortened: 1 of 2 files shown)


nikhilkh:
Could you add a unit test for this?

template/cordova/lib/AppxManifest.js
```diff
@@ -374,6 +374,11 @@ AppxManifest.prototype.getCapabilities = function () {
 };
 
 function refineColor(color) {
+    if (color.indexOf('0x') === -1 && color.indexOf('#') === -1) {
```
dblotsky: Please put this condition into a function; that way it will be easier to debug and read in the future.

dblotsky: Is there a list of valid CSS colours? For example, `meow` and `12345` and are not colours, but they will get accepted here.

nikhilkh: WIndows SDK will do the validation and omit errors for invalid values and it's not for us to validate here.

daserge:
@nikhilkh, @dblotsky, thanks for review, I've addressed the notes.

dblotsky:
LGTM.

PR #143, opened 2016-01-26
CB-10394 Do not cache manifest file while getting package name

vladimir-kotikov:
This fixes [CB-10394](https://issues.apache.org/jira/browse/CB-10394)

### DIFF
--- a/spec/unit/AppxManifest.spec.js
+++ b/spec/unit/AppxManifest.spec.js
@@ -38,6 +38,8 @@ describe('AppxManifest', function () {
         spyOn(xml, 'parseElementtreeSync').andCallFake(function (manifestPath) {
             return XMLS[manifestPath] || parseElementtreeSyncOrig(manifestPath);
         });
+
+        AppxManifest.__set__('manifestCache', {});
     });
 
     describe('constructor', function () {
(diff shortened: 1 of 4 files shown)


nikhilkh:
This design seems to be problematic - From what I understand, there are multiple places where we could be reading and writing a manifest. One of them caches them and hence, can lead to writing stale data. 

One idea is to validate that we never overwrite manifest changes when we write to disk - however, it can detect failure but not recover from it.

nikhilkh:
The fix seems correct and fixes the current symptom - I'm concerned that there are other ways/it's easy to make similar mistakes in the future.

vladimir-kotikov:
@nikhilkh, it seems there is another, more general way to resolve this. We might reuse the same caching facility as used by `ConfigChanges` module. This should resolve the stale cache issue automatically because the one cache will be used for all.

This will require to make some changes in `cordova-common`, but this shouldn't be a problem since we're going to make updates to and release it anyway.

nikhilkh:
I'm OK if we want to fix this tactically right now and take a work item shortly after to fix this correctly.

vladimir-kotikov:
I've created [CB-10463](https://issues.apache.org/jira/browse/CB-10463) for appxmanifest improvements

PR #144, opened 2016-01-27
CB-10446 Windows 10 Cordova Application restart instead of resume

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10446)

### DIFF
--- a/spec/unit/Prepare.Win10.spec.js
+++ b/spec/unit/Prepare.Win10.spec.js
@@ -121,7 +121,11 @@ describe('Windows 10 project', function() {
         applyStartPage(mockConfig.config, mockConfig.manifest, true);
 
         var app = mockConfig.manifest.doc.find('.//Application');
-        expect(app.attrib.StartPage).toBe('ms-appx-web:///www/index.html');
+
+        // Workaround to avoid WWAHost.exe bug: https://issues.apache.org/jira/browse/CB-10446
+        var isAppxWebStartupUri = app.attrib.StartPage === 'ms-appx-web:///www/index.html' || 
+            app.attrib.StartPage === 'ms-appx-web://' + mockConfig.config.packageName() + '/www/index.html';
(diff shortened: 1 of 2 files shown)


nikhilkh:
@rakatyal to review.

nikhilkh:
LGTM

rakatyal:
Looks good. Just a question, do we intend to revert this once the windows team fix it on their side? Is there any harm in including the package name in the URIPrefix for other win platforms as well?

nikhilkh:
I don't see much downside in keeping this always.

daserge:
@nikhilkh, @rakatyal, thank you for the review.
Updated to use `Identity.Name` for 8.1 as well for consistency.

rakatyal:
Thanks @daserge. LGTM.

gasby82:
Hi,
I have the same problem with windows 8.1. My environment is cordova version 6.2.0 and windows platform version 4.3.2. How can i fix it? Any workaround?
Thanks

daserge:
Hi @gasby82, 
It might be caused by this bug: https://issues.apache.org/jira/browse/CB-11443
Try using the master version of the platform as it has been fixed but is not yet released:
`cordova platform rm windows && cordova platform add https://github.com/apache/cordova-windows`

gasby82:
Hi @daserge,
thanks for the reply but with platform version 4.5.0-dev it doesn't work. When I resume my app, it restarts. In Windows 10 it works. Do you know any workaround?

gasby82:
Hi @daserge,
Now it works. Thanks for the support. The problem was due to an incompatibility in the javascript framework. It does not depend on cordova.

daserge:
@gasby82, thanks for the info.
Do you mean `WinJS` by `javascript framework`?

mdudek:
Hi @gasby82, what was the problem with 'javascript framework'?

PR #135, opened 2016-01-06
CB-10303 Fixes build arguments parsing

vladimir-kotikov:
This PR fixes the issues from [CB-10303](https://issues.apache.org/jira/browse/CB-10303)

PR #145, opened 2016-01-28
CB-10303 Add tests to cover arguments parsing regression

vladimir-kotikov:
The PR adds test case for #135


vladimir-kotikov:
@nikhilkh, could you please take a look, when possible

PR #146, opened 2016-01-30
CB-10487 WindowsStoreIdentityName should be lowercased in Application…

daserge:
….StartPage

[Jira issue](https://issues.apache.org/jira/browse/CB-10487)


nikhilkh:
LGTM

PR #147, opened 2016-02-01
CB-10480 Remove Windows 8 template from cordova-windows

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10480)


csantanapr:
For what version is this for? cordova-windows@5.0.0 ?

daserge:
This will be a breaking change, so yes, probably for cordova-windows@5.

purplecabbage:
+1
Windows 8.0 has been deprecated since Apr 13, 2015, let's cleanup so we can provide better focus.

sgrebnov:
Reviewed changes, :+1:

PR #148, opened 2016-02-15
CB-10622 Fix targetsize images being ignored

andorman:
Only "scale" images were properly recognized, targetsize icons were ignored.


sgrebnov:
Hey @andorman, the PR's source and destination branches are incorrect - you can see there are many unrelated commits/changes here.

PR #152, opened 2016-02-18
CB-10640: Windows logging is broken

sarangan12:
**Issue**
In the current system, the logs which are collected in the events must be listed in the CI logs is broken. For example, if there is a syntax error in one of the plugins, mobilespec(during testing) will crash the app. This information is available in event logs. This must be collected in the '[Gathering logs](https://ci.apache.org/builders/cordova-windows-store8.1/builds/1660/steps/gathering-logs/logs/stdio)' step and be available in CI logs. But, that is not happening and must be fixed.  

**Root Cause**
The issue is traced to 2 cause. 

1. **Query** 
The query format that was used to retrieve the event logs is incorrect. For example, even when the 'View Events Logs' tool reports the error, if you run the exact command in command line, it just returns empty results. This query format must be fixed. 

2. **spawn**
The _wevtutil_ command is invoked through spawn (for collecting the event logs). This same approach is used in the same program at 4 different places (for various reasons). In all other scenarios, the output is correct, But, while collecting the event logs, the output is empty. This must be fixed. 

**Fix**
For Issue 1, the query format has been modified to use _timediff_ parameter. This is the approach that is used by the 'View Events Logs' tool. The resulting query is executed on both command line and through the script and confirmed that it is working fine. 

For Issue 2, I suspect that the root cause of the issue might be related to sending the parameters with '/'. But this theory could not be confirmed and modifying this did not work. Instead, the execSync approach did work fine. The same has been used. 

**Testing**
Two scenarios has been tested.

1.  **log using the medic.js**
This approach gets the logs for the last 27 minutes. It produces correct results as expected. 

2.  **Execute log directly without dump option**
In this scenario, it has to continuously display the logs repeatedly with 1 second interval (until manually interrupted).It produces correct results as expected. 

@dblotsky @riknoll @nikhilkh @rakatyal @alsorokin Can you please review and merge this PR?

### DIFF
--- a/template/cordova/lib/log.js
+++ b/template/cordova/lib/log.js
@@ -18,39 +18,42 @@
 */
 
 // requires
-var path         = require('path'),
-    et           = require('elementtree'),
-    ConfigParser = require('./ConfigParser.js'),
-    nopt         = require('nopt');
+var path         = require('path');
+var et           = require('elementtree');
+var ConfigParser = require('./ConfigParser.js');
+var nopt         = require('nopt');
 
-var spawn = require('cordova-common').superspawn.spawn;
+var spawn        = require('cordova-common').superspawn.spawn;
+var execSync     = require('child_process').execSync;
 
 // paths
-var platformRoot = path.join(__dirname, '..', '..'),
-    projectRoot  = path.join(platformRoot, '..', '..'),
-    configPath   = path.join(projectRoot, 'config.xml');
+var platformRoot = path.join(__dirname, '..', '..');
+var projectRoot  = path.join(platformRoot, '..', '..');
+var configPath   = path.join(projectRoot, 'config.xml');
 
 //constants
-var APP_TRACING_LOG = 'Microsoft-Windows-AppHost/ApplicationTracing',
-    ADMIN_LOG       = 'Microsoft-Windows-AppHost/Admin';
+var APP_TRACING_LOG = 'Microsoft-Windows-AppHost/ApplicationTracing';
+var ADMIN_LOG       = 'Microsoft-Windows-AppHost/Admin';
+var ONE_MINUTE      = 60 * 1000;
 
 // variables
-var appTracingInitialState = null,
-    appTracingCurrentState = null,
-    adminInitialState      = null,
-    adminCurrentState      = null,
-    timers                 = [],
-    appName;
+var appTracingInitialState = null;
+var appTracingCurrentState = null;
+var adminInitialState      = null;
+var adminCurrentState      = null;
+var timers                 = [];
+var logFromTime            = 10 * ONE_MINUTE; // show last 10 minutes by default
+var appName;
+
 
 /*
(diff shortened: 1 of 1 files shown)


dblotsky:
1. Why does running the script manually print repeatedly? Is that a use case?
2. What is the 2-second delay for? It doesn't wait on Ctrl-C if the script was run from Bash.
3. Running this using medic silently fails for me: `../cordova-medic/medic/medic.js log --platform windows`.
4. The `Dumping logs starting from 900000` message is not very informative. Is there a way to print a more human-readable timeframe?
5. Are there any tests for this? Specifically, it would be good to verify that the timeouts passed around are correct.

dblotsky:
I just realized that a lot of my above comments apply to the logging script in general, not to this specific fix. This fix seems to work, so it LGTM. Only one question there: it doesn't work unless I run it with Administrator privileges first; is this expected?

alsorokin:
@dblotsky as far as I remember, this script can run without administrator privileges, but you'll need to enable the needed channels first:

`wevtutil set-log "Microsoft-Windows-AppHost/Admin" /e:true /rt:true /ms:4194304`
`wevtutil set-log "Microsoft-Windows-AppHost/ApplicationTracing" /e:true /rt:true /ms:4194304`

https://github.com/apache/cordova-medic/blob/master/SLAVES.md#windows-8-and-windows-81

dblotsky:
@alsorokin yep, I ran those and now it works. LGTM.

dblotsky:
@alsorokin: is there a reason that the code under [this line][line] is even necessary?

[line]: https://github.com/apache/cordova-windows/blob/master/template/cordova/lib/log.js#L68

sarangan12:
+1 to remove the lines. @alsorokin your thoughts?

alsorokin:
The purpose of this code is to remember initial state of the channels to return them back to this state after the script is completed.

If we remove this code, we would leave them open after the script execution.

We can enable/disable the channels only if we have administrative privileges, so the use case of this would be the user monitoring the log output of his program and not our medic script monitoring mobilespec app for the issues.

I think it is worth it to leave this code here because ideally we would want to leave the user's environment in the same state as it was before our script execution.

dblotsky:
@alsorokin that makes sense. Do you think flipping the behavior would be better? That is, to behave as if `--dump` was passed by default, and loop only if a flag (like `--loop`) was passed?

alsorokin:
@dblotsky when I was writing this code, I made continuous logging the default behavior because the log script for Android platform is continuously printing the logcat output.
The ios log command tries to output the log file by using `tail -f` command which should print file contents continuously too.

I think that for the sake of consistency cordova-windows should behave like the other platforms do.

dblotsky:
@alsorokin those other commands print only new messages continuously. The Windows script prints existing messages over and over again, which is a different behavior. We should either make it behave the same, or put the looping behavior behind a flag.

alsorokin:
@dblotsky ouch. That's clearly a bug. It shouldn't do that.

PR #153, opened 2016-02-26
CB-10714 Ignore case for --archs

jerroydmoore:
If "ARM" is passed to --archs, the Appx file is compiled, but the build
will fail due to a case sensitive string comparison.

### DIFF
--- a/template/cordova/lib/build.js
+++ b/template/cordova/lib/build.js
@@ -181,7 +181,7 @@ function parseAndValidateArgs(options) {
     // get build options/defaults
     config.buildType = options.release ? 'release' : 'debug';
 
-    var archs = options.archs || args.archs;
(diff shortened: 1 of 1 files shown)


dblotsky:
@jerroydmoore ping. Sorry, the code fails the CI tests and has been reverted. Please revise and resubmit the PR, and please submit an ICLA (guidelines [here][contributing]).

[contributing]: http://cordova.apache.org/contribute/

purplecabbage:
To be clear, an ICLA is not required, but it is prefered.
This was a one liner, and should not get any ICLA scrutiny.
The failed CI tests however mean this should never have been accepted and merged.

dblotsky:
@jerroydmoore, @purplecabbage is right: that was my mistake. The ICLA is not required, but if you plan to contribute more it would be awesome to get it out of the way sooner.

PR #155, opened 2016-02-29
CB-10730 Move plugin handlers tests into platform's repo

vladimir-kotikov:
JIRA: https://issues.apache.org/jira/browse/CB-10730


spec/unit/JsprojManager.spec.js
```diff
@@ -25,6 +25,7 @@ var AppxManifest = require('../../template/cordova/lib/AppxManifest');
 var JsprojManager = rewire('../../template/cordova/lib/JsprojManager');
 
 var PROJECT_PATH = 'spec/unit/fixtures/EmptyProject';
```
dblotsky: Should this be `DummyProject`?

vladimir-kotikov: Updated

spec/unit/JsprojManager.spec.js
```diff
@@ -53,6 +54,13 @@ describe('JsprojManager', function () {
         JsprojManager.__set__('proj', origProj);
     });
 
+    it('should throw if project is not an windows project', function () {
```
dblotsky: `"... not a windows ..."`

dblotsky:
Is there also a PR for removing these from `cordova-lib`?

vladimir-kotikov:
> Is there also a PR for removing these from cordova-lib?

No, because the code these tests are testing is still in cordova-lib. I think we could remove the tests along with the code as a part of PlatformApi polyfill deprecation (see [CB-10644](https://issues.apache.org/jira/browse/CB-10644) and corresponding [mailing list discussion](http://markmail.org/thread/b5j4mhvnxws7jbgk) )

PR #158, opened 2016-03-10
CB-10838 Use logger from cordova-common

vladimir-kotikov:
This implements https://issues.apache.org/jira/browse/CB-10838


nikhilkh:
LGTM

PR #149, opened 2016-02-16
CB-10622 fix targetsize images being ignored

andorman:

PR #159, opened 2016-03-14
CB-10622 Support any MRT qualifiers for Windows icons

vladimir-kotikov:
JIRA [CB-10622](https://issues.apache.org/jira/browse/CB-10622)

This PR fixes copying MRT images with qualifiers other that `scale-*`
This is an alternative for #149

### DIFF
--- a/spec/unit/Prepare.Win10.spec.js
+++ b/spec/unit/Prepare.Win10.spec.js
@@ -17,6 +17,9 @@
     under the License.
 */
 
+var path = require('path');
+var shell = require('shelljs');
+
 var rewire  = require('rewire'),
     prepare = rewire('../../template/cordova/lib/prepare'),
     AppxManifest = require('../../template/cordova/lib/AppxManifest'),
@@ -123,7 +126,7 @@ describe('Windows 10 project', function() {
         var app = mockConfig.manifest.doc.find('.//Application');
 
         // Workaround to avoid WWAHost.exe bug: https://issues.apache.org/jira/browse/CB-10446
-        var isAppxWebStartupUri = app.attrib.StartPage === 'ms-appx-web:///www/index.html' || 
+        var isAppxWebStartupUri = app.attrib.StartPage === 'ms-appx-web:///www/index.html' ||
             app.attrib.StartPage === 'ms-appx-web://' + mockConfig.config.packageName().toLowerCase() + '/www/index.html';
         expect(isAppxWebStartupUri).toBe(true);
     });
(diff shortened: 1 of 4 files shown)


nikhilkh:
do we need to add scale 200/400 here: https://github.com/apache/cordova-windows/pull/159/files#diff-32bff2189ca335a5581636fff314a504R315

vladimir-kotikov:
@nikhilkh, the items from this array is used only when icon is defined using `width` and `height`. As we decided to recommend using `target` to define icons, should we still continue supporting all available scales for `width`/`height`-defined icons?

spec/unit/Prepare.Win10.spec.js
```diff
@@ -380,3 +383,79 @@ describe('A Windows 10 project should apply the uap: namespace prefix to certain
...
+        expect(warnSpy.calls[0].args[0]).toMatch('image is skipped');
+    });
+
+    describe('when "target" attribute is specified for the image', function () {
+        it('should copy all images with the same base name and extension to destination dir', function () {
+            var matchingFiles = [
+                'res/Windows/Square44x44.scale-100.png',
+                'res/Windows/Square44x44.targetsize-16.png',
+                'res/Windows/Square44x44.scale-150_targetsize-16.png',
+                'res/Windows/Square44x44.targetsize-16_scale-200.png',
```
nikhilkh: Does the qualifier not have to be following a specific syntax for it to work? Should we validate that?

nikhilkh:
Yup - now I understand - we need not need to add to that array.

spec/unit/Prepare.Win10.spec.js
vladimir-kotikov: From [doc](https://msdn.microsoft.com/en-us/library/windows/apps/hh965372.aspx) the syntax is:

> The dotted segment in the file just before the file extension is considered a token that indicates qualifiers. Multiple qualifiers are separated by an underscore, and qualifiers are traditionally specified with their name and value

It's not very clear what happens if the last segment doesn't match any qualifier or has some incorrect qualifier name/value, so i preferred not to perform any validation here.

Also i've just checked and it seems that MSBuild already does some basic validation and emits a warning in this case

```
e:\PROJECTS\Temp\foo
λ cordova build windows
Building project: e:\PROJECTS\Temp\foo\platforms\windows\CordovaApp.Windows.jsproj
        Configuration : debug
        Platform      : anycpu
MakePRI : warning 0xdef00520: Invalid qualifier: DIMENSION-240 [e:\PROJECTS\Temp\foo\platforms\windows\CordovaApp.Windows.jsproj]
MakePRI : warning 0xdef00520: Invalid qualifier: SCALE-FOO [e:\PROJECTS\Temp\foo\platforms\windows\CordovaApp.Windows.jsproj]
MakePRI : warning 0xdef00520: Invalid qualifier: SCALE-240_DIMENSION-FOO [e:\PROJECTS\Temp\foo\platforms\windows\CordovaApp.Windows.jsproj]
```

the build succeeds though, and the resultant package contains improperly named files, so i think we could rely on MSBuild here.

sgrebnov:
Reviewed and verified on sample config/images :+1:

PR #160, opened 2016-03-15
CB-10831 Apply app description to appxmanifest files

vladimir-kotikov:
See [CB-10831](https://issues.apache.org/jira/browse/CB-10831) for background

### DIFF
--- a/template/cordova/lib/prepare.js
+++ b/template/cordova/lib/prepare.js
@@ -172,6 +172,15 @@ function applyCoreProperties(config, manifest) {
         manifest.getVisualElements().setDisplayName(name);
     }
 
+    var description = config.description();
+    manifest.getProperties().setDescription(description);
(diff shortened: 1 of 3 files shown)


template/cordova/lib/prepare.js
```diff
@@ -172,6 +172,15 @@ function applyCoreProperties(config, manifest) {
         manifest.getVisualElements().setDisplayName(name);
     }
 
+    var description = config.description();
+    manifest.getProperties().setDescription(description);
+    if (description) {
```
nikhilkh: This breaks if say you add a description to config.xml and then remove it - the old description will remain in appxmanifest - we might have to revert back the old description in this case.

vladimir-kotikov: Makes, i've updated the PR

PR #162, opened 2016-03-22
CB-10622 Fix targetsize images being ignored

stebi:
In the icon tag in config.xml the icon-"target" attribute is not read, so MRT images are not loaded at all

e.g.
 <icon src="build/windows/storelogo.jpg" target="storelogo" />


vladimir-kotikov:
@stebi, thanks for your contribution, but the issue has been fixed already in dependency module. See apache/cordova-lib#392 The fix will be available in next cordova-windows release

PR #164, opened 2016-03-24
CB-10845 Invalidate manifest cache in prepare

vladimir-kotikov:
This fixes [CB-10845](https://issues.apache.org/jira/browse/CB-10845)

### DIFF
--- a/template/cordova/lib/prepare.js
+++ b/template/cordova/lib/prepare.js
@@ -420,6 +420,11 @@ module.exports.prepare = function (cordovaProject) {
     this._config = updateConfigFilesFrom(cordovaProject.projectConfig,
         this._munger, this.locations);
 
+    // CB-10845 avoid using cached appxmanifests since they could be
+    // previously modififed outside of AppxManifest class
+    // TODO: invalidate only entries that were affected by config munge
+    AppxManifest.purgeCache();
+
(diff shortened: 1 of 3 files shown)


template/cordova/lib/prepare.js
```diff
@@ -420,6 +420,11 @@ module.exports.prepare = function (cordovaProject) {
     this._config = updateConfigFilesFrom(cordovaProject.projectConfig,
         this._munger, this.locations);
 
+    // CB-10845 avoid using cached appxmanifests since they could be
+    // previously modififed outside of AppxManifest class
+    // TODO: invalidate only entries that were affected by config munge
+    AppxManifest.purgeCache();
```
omefire: In this instance, who modifies the file outside the 'AppxManifest' class ?

vladimir-kotikov: It is a `_munger` which applies the changes defined by `<config-file>` elements from `plugin.xml`

omefire:
LGTM!

vladimir-kotikov:
Thanks, @omefire

template/cordova/lib/prepare.js
omefire: Thanks, it makes sense

PR #165, opened 2016-04-07
CB-11024: Add preference to set the min UAP target version in the JSP…

rakatyal:
…roj File

The current preferences allow the user to set the target and min versions in the appxmanifest files. The  target min version in the jsproj file currently is set to be the minimum SDK installed on the user machine. This change allows the user to change it using this new preference else defaults it to '10.0.10240.0'

### DIFF
--- a/template/cordova/lib/ConfigParser.js
+++ b/template/cordova/lib/ConfigParser.js
@@ -73,6 +73,16 @@ WindowsConfigParser.prototype.getWindowsTargetVersion = function() {
     return preference;
 };
 
+WindowsConfigParser.prototype.getUAPTargetMinVersion = function() {
+    var preference = this.getPreference('uap-target-min-version');
+
+    if (!preference) {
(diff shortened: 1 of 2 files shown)


jasongin:
LGTM

template/cordova/lib/ConfigParser.js
```diff
@@ -73,6 +73,15 @@ WindowsConfigParser.prototype.getWindowsTargetVersion = function() {
     return preference;
 };
 
+WindowsConfigParser.prototype.getUAPTargetMinVersion = function() {
+    var preference = this.getPreference('uap-target-min-version');
+
+    if (!preference)
```
nikhilkh: Nit: code style: It just best to always add curlies for blocks - indentation based on editor tabs and other things can sometimes cause errors.

rakatyal:
Merged [here](https://github.com/apache/cordova-windows/commit/07f6158bc24eda28c2946f54b1e3626c7e124c92)

PR #167, opened 2016-04-14
CB-11022 Duplicate www files on plugin installtion/removal

vladimir-kotikov:
This PR is similar to apache/cordova-android#289 and is needed to remove requirement for running prepare after every plugin add/rm

### DIFF
--- a/template/cordova/lib/PluginHandler.js
+++ b/template/cordova/lib/PluginHandler.js
@@ -25,8 +25,8 @@ var shell = require('shelljs');
 var events = require('cordova-common').events;
 var CordovaError = require('cordova-common').CordovaError;
 
-// returns relative file path for a file in the plugin's folder that can be referenced 
-// from a project file. 
+// returns relative file path for a file in the plugin's folder that can be referenced
+// from a project file.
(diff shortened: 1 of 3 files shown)


template/cordova/lib/PluginHandler.js
```diff
@@ -123,24 +123,27 @@ var handlers = {
...
+                removeFile(project.platformWww, path.join('plugins', plugin.id));
+            }
         }
     },
     'js-module': {
         install: function (obj, plugin, project, options) {
             // Copy the plugin's files into the www directory.
             var moduleSource = path.resolve(plugin.dir, obj.src);
-            var moduleName = plugin.id + '.' + (obj.name || path.parse(obj.src).name);
+            var moduleName = plugin.id + '.' + (obj.name || path.basename(obj.src, path.extname (obj.src)));
```
jasongin: Is there a reason for this change? Doesn't it give exactly the same result?

jasongin:
:+1:

template/cordova/lib/PluginHandler.js
vladimir-kotikov: Yup, these are the same, but `path.parse` is not available in Node v0.10

PR #168, opened 2016-04-18
CB-11066 Remove uap prefixed capabilities when plugin has added regular ones

vladimir-kotikov:
This fixes a problem when some of capabilities, added by plugin are being modified at 'prepare' stage due to special Windows logic and then aren't removed when uninstalling plugin. For details see [CB-11066](https://issues.apache.org/jira/browse/CB-11066)

### DIFF
--- a/spec/unit/ConfigChanges.spec.js
+++ b/spec/unit/ConfigChanges.spec.js
@@ -1,0 +1,77 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+var BaseMunger = require('cordova-common').ConfigChanges.PlatformMunger;
+var PlatformMunger = require('../../template/cordova/lib/ConfigChanges').PlatformMunger;
+
+var os = require('os');
+var path = require('path');
+var shell = require('shelljs');
+
+var tempDir = path.join(os.tmpdir(), 'windows');
+var WINDOWS_MANIFEST = 'package.windows.appxmanifest';
+var WINDOWS10_MANIFEST = 'package.windows10.appxmanifest';
+
+describe('PlatformMunger', function () {
+    var munge, munger;
+
+    beforeEach(function () {
+        shell.mkdir('-p', tempDir);
+        munge = { parents: { 'foo/bar': [
+            { before: undefined, count: 1, xml: '<DummyElement name="Dummy" />'}
+        ]}};
+        munger = new PlatformMunger('windows', tempDir);
+        spyOn(BaseMunger.prototype, 'apply_file_munge').andCallThrough();
+    });
+
+    afterEach(function () {
+        shell.rm('-rf', tempDir);
+    });
+
+    describe('apply_file_munge method', function () {
+
+        it('should call parent\'s method with the same parameters', function () {
+            munger.apply_file_munge(WINDOWS_MANIFEST, munge, false);
+            expect(BaseMunger.prototype.apply_file_munge).toHaveBeenCalledWith(WINDOWS_MANIFEST, munge, false);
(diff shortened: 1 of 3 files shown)


nikhilkh:
@rakatyal to help review

template/cordova/Api.js
```diff
@@ -29,7 +29,7 @@ var ActionStack = require('cordova-common').ActionStack;
 var CordovaError = require('cordova-common').CordovaError;
 var CordovaLogger = require('cordova-common').CordovaLogger;
 var PlatformJson = require('cordova-common').PlatformJson;
-var PlatformMunger = require('cordova-common').ConfigChanges.PlatformMunger;
+var PlatformMunger = require('./lib/ConfigChanges').PlatformMunger;
```
rakatyal: Won't removing this cause an issue for the other methods defined in the cordova-common's ConfigChanges class?

template/cordova/lib/ConfigChanges.js
```diff
@@ -0,0 +1,85 @@
...
+ *   need to be removed or added to the file
+ */
+PlatformMunger.prototype.apply_file_munge = function (file, munge, remove) {
+    // Call parent class' method
+    PlatformMunger.super_.prototype.apply_file_munge.call(this, file, munge, remove);
+
+    // CB-11066 If this is a windows10 manifest and we're removing the changes
+    // then we also need to check if there are <Capability> elements were previously
+    // added and schedule removal of corresponding <uap:Capability> elements
+    if (remove && file === 'package.windows10.appxmanifest') {
```
rakatyal: Couldn't we have added this condition to the existing apply_file_munge instead of writing an override?

template/cordova/lib/ConfigChanges.js
```diff
@@ -0,0 +1,85 @@
...
+        var uapCapabilitiesMunge = generateUapCapabilities(munge);
+        PlatformMunger.super_.prototype.apply_file_munge.call(this, file, uapCapabilitiesMunge, remove);
```
rakatyal: Do we need to call this if uapCapabilitiesMunge is empty?

template/cordova/lib/ConfigChanges.js
vladimir-kotikov: This is possible but IMO the idea of cordova-common was to contain a platform-agnostic code, which platforms might reuse in a way that they want to. The disadvantage of putting platform-specific code in common is that we'll need to release common first before releasing platform to make fix available in next version. Keeping platform-specific code directly inside of the platform allows us not to rely on `common` release cycle

template/cordova/lib/ConfigChanges.js
vladimir-kotikov: No, we don't, but since parent's `apply_file_munge` [starts from iterating through munge](https://github.com/apache/cordova-lib/blob/master/cordova-common/src/ConfigChanges/ConfigChanges.js#L78-L79) it is basically a no-op in this case and hence there is no any risks in calling it. Just didn't want to add another check to complicate the code :)

template/cordova/Api.js
vladimir-kotikov: Could you please clarify? I'm not sure how removing the `require` statement might affect required logic.

template/cordova/lib/ConfigChanges.js
rakatyal: Makes sense. Thanks.

template/cordova/lib/ConfigChanges.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+};
+
+/**
+ * Generates a new munge that contains <uap:Capability> elements created based on
+ * corresponding <Capability> elements from base munge. If there is no such elements
+ * found in base munge, the empty munge is returned (selectors might present under
```
rakatyal: If there *are* no such elements ..
selectors might *be* present..

template/cordova/lib/ConfigChanges.js
rakatyal: Sounds fair. Thanks.

spec/unit/ConfigChanges.spec.js
```diff
@@ -0,0 +1,77 @@
...
+
+        it('should remove uap: capabilities added by windows prepare step', function () {
+            // Generate a munge that contain non-prefixed capabilities changes
+            var baseMunge = { parents: { WINDOWS10_MANIFEST: [
+                // Emulate capability that was initially added with uap prefix
+                { before: undefined, count: 1, xml: '<uap:Capability Name=\"privateNetworkClientServer\">'},
+                { before: undefined, count: 1, xml: '<Capability Name=\"enterpriseAuthentication\">'}
+            ]}};
+
+            var capabilitesMunge = { parents: { WINDOWS10_MANIFEST: [
```
rakatyal: Nitpick: capabilitiesMunge

rakatyal:
Minor comments. Otherwise LGTM.

vladimir-kotikov:
Thanks for review, merged

PR #169, opened 2016-04-26
CB-11150 CI Error - Windows Platform - Could not find XHR config file

daserge:
Changes config.xml path to absolute

[Jira issue](https://issues.apache.org/jira/browse/CB-11150)


nikhilkh:
LGTM. Let's get this in - as it will fix our build.

PR #170, opened 2016-04-27
CB-11156 Change default FadeSplashScreenDuration value

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-11156)


rakatyal:
LGTM.

PR #172, opened 2016-04-29
CB-11176 Fix windows-splashscreen compatibility with older plugin ver…

daserge:
…sions

[Jira issue with details](https://issues.apache.org/jira/browse/CB-11176)

### DIFF
--- a/cordova-js-src/exec.js
+++ b/cordova-js-src/exec.js
@@ -41,6 +41,20 @@ var execProxy = require('cordova/exec/proxy');
  */
 module.exports = function (success, fail, service, action, args) {
 
+    // Handle the case when we have an old version of splashscreen plugin to avoid the API calls failures
+    if (service === 'SplashScreen') {
+        var pluginsVersions = require("cordova/plugin_list").metadata;
+        var splashscreenVersion = pluginsVersions['cordova-plugin-splashscreen'];
+        var MIN_SPLASHSCREEN_SUPPORTED_VER = 4;
+        if (splashscreenVersion && ((parseInt(splashscreenVersion.split('.')[0], 10) || 0) < MIN_SPLASHSCREEN_SUPPORTED_VER)) {
+            console.log('Warning: A more recent version of cordova-plugin-splashscreen has been hooked into for compatibility reasons. Update the plugin to version >= 4.');
+
+            var platformSplashscreen = require('cordova/splashscreen');
+            // Replace old plugin proxy with the platform's one
+            require('cordova/exec/proxy').add(service, platformSplashscreen);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


cordova-js-src/exec.js
```diff
@@ -41,6 +41,20 @@ var execProxy = require('cordova/exec/proxy');
  */
 module.exports = function (success, fail, service, action, args) {
 
+    // Handle the case when we have an old version of splashscreen plugin to avoid the API calls failures
+    if (service === 'SplashScreen') {
+        var pluginsVersions = require("cordova/plugin_list").metadata;
+        var splashscreenVersion = pluginsVersions['cordova-plugin-splashscreen'];
+        var MIN_SPLASHSCREEN_SUPPORTED_VER = 4;
+        if (splashscreenVersion && parseInt(splashscreenVersion[0], 10) < MIN_SPLASHSCREEN_SUPPORTED_VER) {
```
rakatyal: Is it supposed to not work for 4.x versions as well?

cordova-js-src/exec.js
```diff
@@ -41,6 +41,20 @@ var execProxy = require('cordova/exec/proxy');
...
+            console.log('Warning: cordova-plugin-splashscreen has been hooked into for compatibility reasons. Update the plugin to version >= 4.');
```
rakatyal: Consider updating the message to something like: "A more recent version of cordova-plugin-splashscreen has been hooked into..."

rakatyal:
Minor comments. Otherwise LGTM.

cordova-js-src/exec.js
daserge: This regards to the plugin version and not to the platform' one,  i.e. it should work starting with cordova-plugin-splashscreen@4 (as it will just redirect API calls to the platform).

cordova-js-src/exec.js
daserge: Updated, thanks.

cordova-js-src/exec.js
```diff
@@ -41,6 +41,20 @@ var execProxy = require('cordova/exec/proxy');
  */
 module.exports = function (success, fail, service, action, args) {
 
+    // Handle the case when we have an old version of splashscreen plugin to avoid the API calls failures
+    if (service === 'SplashScreen') {
+        var pluginsVersions = require("cordova/plugin_list").metadata;
+        var splashscreenVersion = pluginsVersions['cordova-plugin-splashscreen'];
+        var MIN_SPLASHSCREEN_SUPPORTED_VER = 4;
```
cadesalaberry: From this line, should I understand that the next 4.x.x release here https://github.com/apache/cordova-plugin-splashscreen/releases will be fully compatible with windows ?

PR #173, opened 2016-04-29
CB-10653 Making the windows activation context complete

rakatyal:
The current implementation doesn't capture details for all the activation options for UWP apps.

### DIFF
--- a/cordova-js-src/platform.js
+++ b/cordova-js-src/platform.js
@@ -27,7 +27,8 @@ module.exports = {
             channel = cordova.require('cordova/channel'),
             platform = require('cordova/platform'),
             modulemapper = require('cordova/modulemapper'),
-            configHelper = require('cordova/confighelper');
+            configHelper = require('cordova/confighelper'),
+            utils = require('cordova/utils');
 
         modulemapper.clobbers('cordova/exec/proxy', 'cordova.commandProxy');
 
@@ -56,9 +57,9 @@ module.exports = {
             // OR cordova.require('cordova/platform').activationContext
             // activationContext:{type: actType, args: args};
             var activationHandler = function (e) {
-                var args = e.detail.arguments;
-                var actType = e.detail.type;
-                platform.activationContext = { type: actType, args: args };
+                // Making all the details available as activationContext
+                platform.activationContext = utils.clone(e.detail); 
+                platform.activationContext.args = e.detail.arguments;       /* for backwards compatibility */
 
                 function makePromise(fn) {
                     return new WinJS.Promise(function init(completeDispatch, errorDispatch) {
--- a/template/www/cordova.js
+++ b/template/www/cordova.js
@@ -1479,7 +1479,8 @@ module.exports = {
             channel = cordova.require('cordova/channel'),
             platform = require('cordova/platform'),
             modulemapper = require('cordova/modulemapper'),
-            configHelper = require('cordova/confighelper');
+            configHelper = require('cordova/confighelper'),
+            utils = require('cordova/utils');
 
         modulemapper.clobbers('cordova/exec/proxy', 'cordova.commandProxy');
 
@@ -1508,9 +1509,9 @@ module.exports = {
             // OR cordova.require('cordova/platform').activationContext
             // activationContext:{type: actType, args: args};
             var activationHandler = function (e) {
-                var args = e.detail.arguments;
-                var actType = e.detail.type;
-                platform.activationContext = { type: actType, args: args };
+                // Making all the details available as activationContext
+                platform.activationContext = utils.clone(e.detail); 
+                platform.activationContext.args = e.detail.arguments;       /* for backwards compatibility */
 
                 function makePromise(fn) {
                     return new WinJS.Promise(function init(completeDispatch, errorDispatch) {


riknoll:
Was there a reason we weren't saving all of the detail information before?

rakatyal:
Not sure about that. May be it wasn't needed at that time since the activation events were limited. With UWP there are more activation kinds introduced.

nikhilkh:
@purplecabbage who was involved in some of the original code here. @daserge @vladimir-kotikov to review as well

cordova-js-src/platform.js
```diff
@@ -56,9 +57,10 @@ module.exports = {
             // OR cordova.require('cordova/platform').activationContext
             // activationContext:{type: actType, args: args};
             var activationHandler = function (e) {
-                var args = e.detail.arguments;
-                var actType = e.detail.type;
-                platform.activationContext = { type: actType, args: args };
+                // Making all the details available as activationContext
+                platform.activationContext = utils.clone(e.detail); 
+                platform.activationContext.args = e.detail.arguments;       /* for backwards compatibility */
+                platform.activationContext.actType = e.detail.type;         /* for backwards compatibility */
```
jasongin: I think you don't need an actType property for back-compat. The property (not the variable) was previously named just "type".

template/www/cordova.js
```diff
@@ -1508,9 +1509,10 @@ module.exports = {
             // OR cordova.require('cordova/platform').activationContext
             // activationContext:{type: actType, args: args};
             var activationHandler = function (e) {
-                var args = e.detail.arguments;
-                var actType = e.detail.type;
-                platform.activationContext = { type: actType, args: args };
+                // Making all the details available as activationContext
+                platform.activationContext = utils.clone(e.detail); 
+                platform.activationContext.args = e.detail.arguments;       /* for backwards compatibility */
+                platform.activationContext.actType = e.detail.type;         /* for backwards compatibility */
```
jasongin: Same here with actType.

cordova-js-src/platform.js
rakatyal: That's right! Corrected.

jasongin:
👍

cordova-js-src/platform.js
```diff
@@ -56,9 +57,9 @@ module.exports = {
             // OR cordova.require('cordova/platform').activationContext
             // activationContext:{type: actType, args: args};
             var activationHandler = function (e) {
-                var args = e.detail.arguments;
-                var actType = e.detail.type;
-                platform.activationContext = { type: actType, args: args };
+                // Making all the details available as activationContext
+                platform.activationContext = utils.clone(e.detail); 
+                platform.activationContext.args = e.detail.arguments;       /* for backwards compatibility */
```
purplecabbage: Is it intentional that we pass the actual object in as `args`?
Is that the backwards compatibility part, or is it just that the object has a property `args` instead of `arguments`
If so I would prefer `platform.activationContext.args = platform.activationContext.arguments;`

rakatyal: Yes that is just for the backwards compatibility since I assume plugins would already be using activationContext.args property. Would changing it to `platform.activationContext.args = platform.activationContext.arguments` make any difference?

purplecabbage: I just mean that we have gone thru the trouble of using utils.clone, which to me implies that we want a shallow copy, and then we were also passing a deep copy of arguments.

ogoguel:
+1 for that feature!

A quick question though :how to pass the `platform.activationContext` back as a  `WebUILaunchActivatedEventArgs` object to a C# Windows Runtime Component  ?

Ex: ` public static void onActivated(WebUILaunchActivatedEventArgs _event)`

I would assume that if it has been cloned, the object class is being lost, but the `WebUILaunchActivatedEventArgs `is not constructible.

rakatyal:
@ogoguel: I guess I can pass the actual object as it is to platform.activationContext instead of cloning it. That should solve your use-case?

cordova-js-src/platform.js
rakatyal: That's true. Users might need to modify the object so I guess I will pass the actual object instead of cloning it.

afourney:
I was commenting on JIRA, sorry.

I definitely needed to clone it when I submit the initial request. I spent a good amount of time debugging this issue.

I guess that e.detail.arguments is stable, since this is what the existing code uses. But e.detail.result would not persist outside of the handler. This is why I used clone.

The point about mixing strategies is well taken. The patch probably should be:
platform.activationContext.args = platform.activationContext.arguments; // for backward compatibility

afourney:
Also, my primary use case is launching an app via Cortana. (e.g., https://code.msdn.microsoft.com/windowsapps/Voice-Commands-Quickstart-64ce68ce/sourcecode?fileId=124843&pathId=1713588077)

Importantly, the recognized intent, semantic details, and results from the speech recognizer, are accessed through: e.detail.result

This wasn't getting saved, making it impossible to develop a full-featured Cortana plugin.

rakatyal:
@afourney: Thanks for the info. I am a little skeptical of changing the way args were being passed since people might have been modifying the original object. To keep a uniform strategy, I would pass the actual detail object too instead of cloning it.

afourney:
@rakatyal Understood. Tomorrow I'll try to reproduce what I was seeing before. In any case, detail.result *must* be cloned. The result object is undefined outside of the activation handler, and, in particular,  cannot by the time Cordova's deviceready triggers.

rakatyal:
@afourney: Thanks. You should be able to access it with this change by platform.activationContext.detail.result.

ogoguel:
@rakatyal, actually, I believe you need to clone it anyway to have it available in the trigger.

My problem is more about how to serialize/marshaling such an `WebUILaunchActivatedEventArgs` parameter so it can be used in C#, but this is maybe off topic...

Looking forward to have this PR merged!

rakatyal:
I did some digging into this with other activation kinds and I confirmed that few event details are lost once the event handler is executed possibly to hide/disable few properties of the object. So it looks like cloning might be the only way out of it. I will update the PR soon.

afourney:
@rakatyal Thanks for investigating. Your findings agree with my own experiences. This is possibly some quirk of WinJS.

rakatyal:
Thanks for the feedback guys. Merging the changes.

cnsdato:
What was the reason for cloning the detail object?
I have 2 issues when kind is Windows.ApplicationModel.Activation.ActivationKind.file for file association:

1. The clone fails because neighboringFilesQuery.folder property is not accessible
2. When trying to read the StorageFile via Windows.Storage.FileIO.readTextAsync() I get a type mismatch.

afourney:
The object is cloned because some of its fields appear only to be available when inside the event handler. By the time one's Cordova script can access platform.activationContext, important details are lost (i.e., the activationContext would otherwise change underfoot).

@cnsdato 
Clearly cloning failing is a problem and will need to be addressed. Also, clearly the original code (that only saved the command line arguments) would not serve your purpose.  Have you tested if you are able to do what you need if clone is removed? .

cnsdato:
@afourney
I have tested it without cloning the object and it works, in my case of `Windows.ApplicationModel.Activation.ActivationKind.file`

daserge:
@cnsdato, could you please file a corresponding Jira issue at http://issues.cordova.io/?

cnsdato:
@daserge, issue was created: https://issues.apache.org/jira/browse/CB-11522

daserge:
@cnsdato, thank you!

daserge:
@rakatyal, @afourney, do you happen to have some examples with details disappearing?
I tried and it did not reproduce for me on Windows Phone 8.1 project running on Windows 10 Mobile.
Which OS and project types was it reproducing on for you BTW?

daserge:
@rakatyal, you also noted that it was happening for some other activation kinds - could you please share which ones I should also test?
What do you think on a solution to have the raw args.details AND the cloned object so that a developer could make a call to the clone if some property is missing. This could help with types loss as reported [here](https://github.com/apache/cordova-windows/pull/173#issuecomment-227366714).

rakatyal:
@dasarge: I don't recall the exact OS but I could reproduce @afourney's problem. The details were lost by the time I could access it in plugin's code. I think I could see this happening for FileSavePicker kind as well. As far as having both the raw and cloned details object, I am fine with it.

daserge:
@rakatyal, thanks for the details!

afourney:
To follow up: I was testing on Windows 10 desktop and Windows 10 mobile. On both platforms the e.detail.result field goes missing when e.detail.kind === activation.ActivationKind.voiceCommand and accessed outside of the event handler.

I don't think that both cloning and keeping the original object is a solution. It sounds like clone fails when it's Activation.ActivationKind.file.  I suppose we could wrap the clone in a try/catch, or build a more robust clone operator for this use case.

daserge:
@afourney, see https://issues.apache.org/jira/browse/CB-11522?focusedCommentId=15359160&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15359160 - we'll need to fix the clone method too, but it will not help missing the native types of cloned properties like StorageFile (so that a successive call to `Windows.Storage.FileIO.readTextAsync` will fail, for this example).
To confirm - were you testing on an UWP project?

afourney:
Yes. It was all UWP and obviously all with WinJS.

PR #174, opened 2016-05-03
CB-11117: Optimize prepare for windows platform, clean prepared files

jasongin:
This uses the `FileUpdater` module added in apache/cordova-lib#429 to skip copying files that didn't change. Some refactoring was required because previously the target directories would just be wiped before copying; now we need to map out the source and target directories so the `FileUpdater` has the necessary information to determine the optimal set of file operations.

And I'm making it so that `cordova clean windows` also cleans the prepared files (but `clean.bat` does not).


nikhilkh:
Looks like there are some unit test failures

jasongin:
Oh, yeah I'll update those tests.

---
In reply to: [218305527](https://github.com/apache/cordova-windows/pull/174#issuecomment-218305527) [](ancestors = 218305527)

template/cordova/lib/prepare.js
```diff
@@ -425,19 +451,39 @@ module.exports.prepare = function (cordovaProject) {
...
     });
 };
 
+module.exports.clean = function (options) {
+    // A cordovaProject isn't passed into the clean() function, because it might have
+    // been called from the platform shell script rather than the CLI. Check for the
+    // noPrepare option passed in by the non-CLI clean script. If that's present, or if
+    // there's no config.xml found at the project root, then don't clean prepared files.
+    var projectRoot = path.resolve(this.root, '../..');
+    var projectConfigFile = path.join(projectRoot, 'config.xml');
```
nikhilkh: @vladimir-kotikov  to help review - as far as I remember platforms never parse the root config. In particular, the root config can have conditional elements - 
```xml
<platform name="ios"> 
    <icon ... > 
</platform>
```

nikhilkh: I see you are invoking `getIcons('windows')` and this should do the right thing. Is there a reason why you do not use the config.xml in the platform directory?

jasongin: That's a reasonable question. I didn't want the clean function to depend on a previous prepare having completed successfully (so that the platform config.xml was merged). But I'm not sure that's best, and maybe it would be cleaner just to use that file.

I did this the same way for the other 2 platforms, so this discussion would apply there also.

jasongin: I updated the PRs for all 3 platforms to use the platform's config.xml for cleaning.

They still check for the presence of config.xml at the project root, because I want to be extra careful not to do this clean operation in a platform-centric workflow.

vladimir-kotikov:
LGTM

PR #175, opened 2016-05-09
Fix build result when using --bundle and multiple architectures

georgeslegros:
Hello,

This small modification aims at avoiding error messages when building appxbundle with multiple architectures.

The previous code was not taking archs into account.

Cheers,


CSharpFan:
`npm install lodash --save`

nikhilkh:
@georgeslegros Thanks for this fix!

It will be great to not add a dependency to lodash for this - also as you have not added it to package.json - our tests are failing.

It will be great if you can create a JIRA here about the issue with details on how to reproduce it - this will help in testing this fix: https://issues.apache.org/. Please follow instructions here on how to file a JIRA: http://cordova.apache.org/contribute/issues.html

PR #177, opened 2016-05-13
CB-11204: Catch when SDK not present on build and provide appropriate…

rakatyal:
… message

Currently we don't have any check before calling the build through msbuild which doesn't give a meaningful error when the required SDK is not installed. We should catch this before and respond with the appropriate error message.

### DIFF
--- a/bin/lib/check_reqs.js
+++ b/bin/lib/check_reqs.js
@@ -62,21 +62,7 @@ var REQUIRED_VERSIONS = {
     }
 };
 
-function getConfig() {
-    try {
-        return new ConfigParser(path.join(__dirname, '../../config.xml'));
-    } catch (e) {
-        throw new CordovaError('Can\'t check requirements for Windows platform.' +
-            'The config.xml file is either missing or malformed.');
-    }
-}
-
-
-function getMinimalRequiredVersionFor (requirement) {
-
-    var config = getConfig();
-    var windowsTargetVersion = config.getWindowsTargetVersion();
-    var windowsPhoneTargetVersion = config.getWindowsPhoneTargetVersion();
+function getMinimalRequiredVersionFor (requirement, windowsTargetVersion, windowsPhoneTargetVersion) {
 
     if (windowsTargetVersion === '8' || windowsTargetVersion === '8.0') {
         throw new CordovaError('windows8 platform is deprecated. To use windows-target-version=8.0 you may downgrade to cordova-windows@4.');
@@ -85,7 +71,6 @@ function getMinimalRequiredVersionFor (requirement) {
     if (windowsPhoneTargetVersion === '8' || windowsPhoneTargetVersion === '8.0') {
         throw new CordovaError('8.0 is not a valid version for windows-phone-target-version (use the wp8 Cordova platform instead)');
     }
-
     var windowsReqVersion = Version.tryParse(REQUIRED_VERSIONS[windowsTargetVersion][requirement]);
     var phoneReqVersion = Version.tryParse(REQUIRED_VERSIONS[windowsPhoneTargetVersion][requirement]);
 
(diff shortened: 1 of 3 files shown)


nikhilkh:
@vladimir-kotikov to help review

bin/lib/check_reqs.js
```diff
@@ -362,6 +362,36 @@ module.exports.run = function () {
...
+/** Checks if Windows SDK required to build the target_platform is present
+ * @param {String}  target_platorm        Target platform ('8.1' or '10.0')
+ * @return {Boolean}                      Whether the min SDK required to build the platform is present  
+ *
+ */ 
+module.exports.isWinSDKPresent = function (target_platform) {
+    return getInstalledWindowsSdks()
+    .then(function (installedSdks) {
+        var reqVersion = Version.tryParse(REQUIRED_VERSIONS[target_platform].windowssdk);
+        var hasSdkInstalled = installedSdks.some(function (installedSdk) {
```
omefire: It seems like ```getInstalledWindowsSdks()``` doesn't return an array in all cases, i.e: when an exception occurs while running the registry query. Maybe, modify it to return an empty array in such a case, otherwise we'll error out here.

omefire: same goes for ```isPhoneSDKPresent()```

vladimir-kotikov: In `getInstalledWindowsSdks` If registry search fails, `fail` handler passes empty string below, as if it didn't get any output. Due to this regex search fails to find any SDKs and the last line returns an empty array
```js
     var installedSdks = [];
     // ... skipped ...
     .thenResolve(installedSdks);
```

bin/lib/check_reqs.js
```diff
@@ -362,6 +362,36 @@ module.exports.run = function () {
...
+            return installedSdk.eq(reqVersion);
+        });
+
+        return hasSdkInstalled ? true : false;
```
vladimir-kotikov: This line is excess. `return installedSdks.some(function (installedSdk) {...` should be sufficient

bin/lib/check_reqs.js
```diff
@@ -362,6 +362,36 @@ module.exports.run = function () {
...
+// Checks if min SDK required to build Windows Phone 8.1 project is present
+module.exports.isPhoneSDKPresent = function () {
+    return getInstalledPhoneSdks()
+    .then(function (installedSdks) {
+        var reqVersion = Version.tryParse(REQUIRED_VERSIONS['8.1'].phonesdk);
+        var hasSdkInstalled = installedSdks.some(function (installedSdk) {
+            return installedSdk.eq(reqVersion);
+        });
+
+        return hasSdkInstalled ? true : false;
```
vladimir-kotikov: Same here

template/cordova/lib/MSBuildTools.js
```diff
@@ -44,8 +52,39 @@ MSBuildTools.prototype.buildProject = function(projFile, buildType, buildarch, o
             args.push('/p:' + key + '=' + otherConfigProperties[key]);
         });
     }
+    var that = this;
 
-    return spawn(path.join(this.path, 'msbuild'), [projFile].concat(args), { stdio: 'inherit' });
+    // Check if SDK required to build the respective platform is present. If not present, return with corresponding error, else call msbuild.
+    if (projFile.indexOf('CordovaApp.Phone.jsproj') > -1) {
```
vladimir-kotikov: These 3 blocks below are almost identical. It could more convenient if `checkPhoneSDK` and `checkWinSDK` would return a promise, rejected with corresponding error message if SDK is not found. Then you could just call `check` method and call `spawn` on success.

bin/lib/check_reqs.js
omefire: Hmm, so no early exits ? Interesting.

bin/lib/check_reqs.js
rakatyal: +1

template/cordova/lib/MSBuildTools.js
rakatyal: I implemented it earlier that way but that didn't really look any cleaner (still required 3 different function calls, 3 different error messages and spawn calls). I thought this would look more readable. Thoughts?

vladimir-kotikov: I thought about something like that (assuming that `check*SDK` will throw if SDK not found):
```
var promise;

if (projFile.indexOf('CordovaApp.Phone.jsproj') > -1) {
    promise = checkPhoneSDK();
} else if (projFile.indexOf('CordovaApp.Windows.jsproj') > -1) {
    promise = checkWinSDK('8.1');
} else {
    promise = checkWinSDK('10.0');
}

return promise.then(function () {
    return spawn(path.join(that.path, 'msbuild'), [projFile].concat(args), { stdio: 'inherit' });
})
```

vladimir-kotikov: Anyway, this is just a matter of taste, and i'm fine with your approach.

rakatyal:
@nikhilkh , @vladimir-kotikov : Refactored the code to be more compact. Please have a look.

vladimir-kotikov:
LGTM

nikhilkh:
LGTM. Thanks, Raghav for doing the refactor!

PR #178, opened 2016-06-15
CB-11443 Splashscreen is created the second time on resume on Windows

daserge:
Jira issue: https://issues.apache.org/jira/browse/CB-11443

Looks like github repo is out of sync with Apache.


vladimir-kotikov:
LGTM

PR #179, opened 2016-06-30
CB-11516 windows: Preparing icons w/ target fails

hoovercj:
Putting the following line in the config will will cause an error: "Source
directory does not exist: Square44x44Logo_100.scale"
<icon src="res/Windows/Square44x44Logo_100.png" target="SmallIcon"/>
In prepare.js, `copyImages` calls `mapImageResources` but subdirectory
information was lost. `copyMrtImage` in an earlier commit preserved this
information (shown in the link below), but it was lost in a refactoring.
https://github.com/apache/cordova-windows/blob/c1b80be51cdbd90160b6cb8162350bced8e1c44b/template/cordova/lib/prepare.js#L316
This was missed by a defect in the `copyIcons` test in
Prepare.Win10.spec.js:
`readdirSync` was spied on to return an array of file paths when it
actually only returns an array of the filenames. This caused the
`copyIcons` spec to pass but the code failed when used.

### DIFF
--- a/spec/unit/Prepare.Win10.spec.js
+++ b/spec/unit/Prepare.Win10.spec.js
@@ -507,16 +507,16 @@ describe('copyIcons method', function () {
     describe('when "target" attribute is specified for the image', function () {
         it('should copy all images with the same base name and extension to destination dir', function () {
             var matchingFiles = [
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
LGTM

PR #180, opened 2016-06-30
CB-11470 App crashes when trying to open from another app using Custo…

daserge:
…m URL (Protocol)

[Jira issue](https://issues.apache.org/jira/browse/CB-11470)

According to [the docs](https://msdn.microsoft.com/library/windows/apps/br224763):
> The [SplashScreen](https://msdn.microsoft.com/en-us/library/windows/apps/windows.applicationmodel.activation.iactivatedeventargs.splashscreen) property is available on all [Windows.ApplicationModel.Activation](https://msdn.microsoft.com/en-us/library/windows/apps/br229866) classes.

### DIFF
--- a/cordova-js-src/splashscreen.js
+++ b/cordova-js-src/splashscreen.js
@@ -286,20 +286,18 @@ var splash = null; // Variable to hold the splash screen object.
 var coordinates = { x: 0, y: 0, width: 0, height: 0 }; // Object to store splash screen image coordinates. It will be initialized during activation. 
 
 function activated(eventObject) {
-    if (eventObject.detail.kind === Windows.ApplicationModel.Activation.ActivationKind.launch) {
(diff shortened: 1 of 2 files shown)


purplecabbage:
What is the type of Windows.ApplicationModel.Activation.ActivationKind?

daserge:
It's [Windows.ApplicationModel.Activation.ActivationKind](https://msdn.microsoft.com/en-us/library/windows/apps/windows.applicationmodel.activation.activationkind.aspx).
The gist is that we are calling `splashscreen.firstShow` and thus the `activated` function only when [ApplicationExecutionState not in (running, suspended)](https://github.com/daserge/cordova-windows/blob/37f6494beef23b88bcf91420aff7e8e7c39a328f/cordova-js-src/platform.js#L74), i.e. when the splashscreen is not initialized.

PR #185, opened 2016-07-12
CB-11558 Make windows plugin rm remove ProjectReference items

daserge:
The issue was that we have been passing a relative path during `JsprojManager`.`addProjectReference` and an absolute path during `JsprojManager`.`removeProjectReference` so that regexp failed to find and remove it.
The tests were passing as they were also using absolute paths to the dummy plugin (it was not being added to the test project at all).

Initial idea was to use relative paths instead for project sharing: [CB-10927 Framework references in plugin.xml file prevent Windows Universal projects from being used on other machines](https://issues.apache.org/jira/browse/CB-10927)

Also updated the tests to use relative paths to the ProjectReferences.

[Jira issue](https://issues.apache.org/jira/browse/CB-11558)

### DIFF
--- a/template/cordova/lib/PluginHandler.js
+++ b/template/cordova/lib/PluginHandler.js
@@ -88,7 +88,7 @@ var handlers = {
             var targetDir = obj.targetDir || '';
 
             if(type === 'projectReference') {
-                dest = path.join('plugins', plugin.id, targetDir, src);
+                dest = path.join(path.relative(project.projectFolder, plugin.dir), targetDir, src);
                 project.addProjectReference(dest, getTargetConditions(obj));
             } else {
                 // path.join ignores empty paths passed so we don't check whether targetDir is not empty
@@ -105,7 +105,7 @@ var handlers = {
             var type = obj.type;
 
             if(type === 'projectReference') {
-                project.removeProjectReference(path.join(plugin.dir, src), getTargetConditions(obj));
+                project.removeProjectReference(path.join(path.relative(project.projectFolder, plugin.dir), src), getTargetConditions(obj));
             }
             else {
                 var targetPath = path.join('plugins', plugin.id);
--- a/spec/unit/pluginHandler/windows.spec.js
+++ b/spec/unit/pluginHandler/windows.spec.js
@@ -65,7 +65,8 @@ beforeEach(function () {
             };
 
             return xml.find(xpath) !== null;
-        }    });
+        }
+    });
 });
 
 var getPluginFilePath = PluginHandler.__get__('getPluginFilePath');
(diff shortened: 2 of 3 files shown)


template/cordova/lib/PluginHandler.js
```diff
@@ -105,7 +105,7 @@ var handlers = {
             var type = obj.type;
 
             if(type === 'projectReference') {
-                project.removeProjectReference(path.join(plugin.dir, src), getTargetConditions(obj));
+                project.removeProjectReference(path.join('..', '..', 'plugins', plugin.id, src), getTargetConditions(obj));
```
vladimir-kotikov: @daserge, please don't use hardcoded `../../plugins` path here. The idea here is that PlatformApi abstracts from cordova project structure so we ideally should not rely on presence of `plugins` directory two levels above. I'd rather calculate relative path from `plugin.dir` using `path.relative`

vladimir-kotikov: Another thing is that when we add project reference, we use [`targetDir`](https://github.com/apache/cordova-windows/pull/185/files#diff-7bcd02ec7f03ffbb25512bd1f4a0d388L91) attribute when defined. Shouldn't we add it to path here as well?

daserge: Regarding `targetDir` - this is [not supported](https://github.com/daserge/cordova-windows/blob/ee570036bc8faddc7d63e2830144c3d007d8e904/spec/unit/pluginHandler/windows.spec.js#L140) for projectReferences.
Regarding hardcoded `../..` - using `path.join(path.relative(project.projectFolder, plugin.dir), src)` instead is causing some issues with tests related to project and plugin different drives location, investigating.

vladimir-kotikov: > Regarding targetDir - this is not supported for projectReferences.

I see, thanks

daserge: Updated, thanks for review!
There was an issue with mock dummyplugin, which PluginInfo.dir was initialized with a corresponding fixture path - so in case when the temp folder was on a different drive it caused path.relative to resolve in an absolute path in [PluginHandler.framework.install](https://github.com/daserge/cordova-windows/blob/2bc5d96bdd0f6c9328680bede1b535dde218e23c/template/cordova/lib/PluginHandler.js#L91).
Normally we have a plugin.dir path initialized with relative path to the project root, i.e. `project_root/plugins/plugin-id` - so I've updated the tests accordingly.
I've also tested that install/rm a plugin from another drive updates jsproj files properly.

vladimir-kotikov:
LGTM, :ship: it

PR #186, opened 2016-07-15
Fix duplicate capability declarations in Win10 AppxManifest

rs22:
I have created an issue for this in Jira: https://issues.apache.org/jira/browse/CB-11582

Let me describe the issue in more detail - I have this .\testplugin\plugin.xml:
```
<?xml version='1.0' encoding='utf-8'?>
<plugin id="testplugin" version="1" xmlns="http://apache.org/cordova/ns/plugins/1.0">
    <name>testplugin</name>
    
    <platform name="windows">
        <config-file parent="/Package/Capabilities" target="package.windows10.appxmanifest">
            <Capability Name="enterpriseAuthentication" />
            <Capability Name="sharedUserCertificates" />
        </config-file>
    </platform>
</plugin>
```

When I run `cordova create app && pushd app && cordova plugin add ..\testplugin && cordova platform add windows` the package.windows10.appxmanifest will have the capabilities set correctly. Now, everytime I run `cordova prepare`, this will duplicate all the uap-prefixed capabilities:

```
    <Capabilities>
        <Capability Name="internetClient" />
        <uap:Capability Name="enterpriseAuthentication" />
        <uap:Capability Name="sharedUserCertificates" />
        <uap:Capability Name="enterpriseAuthentication" />
        <uap:Capability Name="sharedUserCertificates" />
    </Capabilities>
```

I had a look at how the capability declarations are handled in ConfigChanges.js and AppxManifest.js. The easiest fix seems to be to just check for duplicated capabilities when writing the AppxManifest (which is what I've implemented). Another way could be to inject the uap prefixes before applying the ConfigChanges, so that the capabilities are not being duplicated in the first place.

Let me know what you think and I'll update the code accordingly.

### DIFF
--- a/spec/unit/AppxManifest.spec.js
+++ b/spec/unit/AppxManifest.spec.js
@@ -25,6 +25,7 @@ var Win10AppxManifest = AppxManifest.__get__('Win10AppxManifest');
 var refineColor = AppxManifest.__get__('refineColor');
 
 var WINDOWS_MANIFEST = 'template/package.windows.appxmanifest';
+var WINDOWS_10_MANIFEST = 'template/package.windows10.appxmanifest';
 var WINDOWS_PHONE_MANIFEST = 'template/package.phone.appxmanifest';
 var CSS_COLOR_NAME = 'turquoise';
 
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
@rs22, thank you for your contribution. The patch looks good, but before merging could you please file Apache CLA as explained [here](https://github.com/apache/cordova-windows/blob/master/CONTRIBUTING.md) and let me know when it accepted. Thanks in advance.

rs22:
@vladimir-kotikov I've submitted my CLA today and it has already been approved.

vladimir-kotikov:
Done.

PR #187, opened 2016-07-19
CB-11241 Return adding BOM to www back to prepare

vladimir-kotikov:
This PR reverts changes, made in #134

The reason of reverting is that original change causes the problems with submitting apps to Windows store when app is built using Visual Studio, because BOM signatures were being added by cordova-windows in `build.js`. In this PR processing `www` is moved back to prepare, and hence all `www` assets will be properly updated to have BOM when the project is opened in VS.

For more background see JIRA issue [CB-11241](https://issues.apache.org/jira/browse/CB-11241)

### DIFF
--- a/template/cordova/lib/prepare.js
+++ b/template/cordova/lib/prepare.js
@@ -458,6 +458,9 @@ module.exports.prepare = function (cordovaProject, options) {
     })
     .then(function () {
         copyImages(cordovaProject, self.locations);
+        // CB-5421 Add BOM to all html, js, css files
+        // to ensure app can pass Windows Store Certification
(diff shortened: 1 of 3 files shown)


sgrebnov:
LGTM :+1:

template/cordova/lib/prepare.js
```diff
@@ -512,8 +515,6 @@ function addBOMSignature(directory) {
     });
```
purplecabbage: +1

purplecabbage:
+1 :shipit:

toddm:
:shipit:

PR #188, opened 2016-07-21
CB-11548 Fix issues where MSBuild cannot be found

jicongw:
…e. Let Cordova Windows look for MSBuild under VSINSTALLDIR process environment variable. Also, make Cordova Windows more resilient to MSBuild version change. This commit contains work by Vladimir and Jason. Code is reviewed by Vladimir and Tim Barham. Commits are squashed.

### DIFF
--- a/template/cordova/lib/MSBuildTools.js
+++ b/template/cordova/lib/MSBuildTools.js
@@ -84,7 +84,7 @@ module.exports.findAvailableVersion = function () {
     });
 };
 
-module.exports.findAllAvailableVersions = function () {
+function findAllAvailableVersionsFallBack() {
     var versions = ['15.0', '14.0', '12.0', '4.0'];
     events.emit('verbose', 'Searching for available MSBuild versions...');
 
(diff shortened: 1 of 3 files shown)


template/cordova/lib/MSBuildTools.js
```diff
@@ -93,6 +93,20 @@ module.exports.findAllAvailableVersions = function () {
             return !!item;
         });
     });
+}
+
+module.exports.findAllAvailableVersions = function () {
+    // CB-11548 use VSINSTALLDIR environment if defined to find MSBuild. If VSINSTALLDIR
+    // is not specified or incorrect - fall back to default discovery mechanism.
```
TimBarham: It is quite likely that `VSINSTALLDIR` is specified and *correct*, but just doesn't contain the msbuild path we're looking for (this will be the case for previous versions of VS, for example). So I think it would be clearer to say here something like "not specified or doesn't contain the MSBuild path we are looking for".

TimBarham:
Thanks @jicongw - I just had one comment, plus fixing the commit message as I mentioned offline.

TimBarham:
Merged in 2d7d219fc1864fa22f87cebadac173fd41091351.

Sorry @jicongw, I forgot to tag this PR in the modified commit, so you will have to close the PR manually.

jicongw:
This has been merged.

PR #190, opened 2016-08-02
CB-11478 Parse --archs option consistently

vladimir-kotikov:
This PR aligns arguments parsing logic with `build` command where passing `--archs` either as CLI argument (`cordova build --archs`) or platform arg, behind `--` (`cordova build -- --archs`) doesn't make any difference.

See also original issue: [CB-11478](https://issues.apache.org/jira/browse/CB-11478)

### DIFF
--- a/template/cordova/lib/run.js
+++ b/template/cordova/lib/run.js
@@ -39,7 +39,7 @@ module.exports.run = function (options) {
 
     // parse arg
     var args  = nopt({
-        'archs': String,
+        'archs': [String],
         'phone': Boolean,
         'win': Boolean,
         'appx': String,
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
@daserge, @sgrebnov, please review

template/cordova/lib/run.js
```diff
@@ -39,7 +39,7 @@ module.exports.run = function (options) {
 
     // parse arg
     var args  = nopt({
-        'archs': String,
+        'archs': [String],
```
daserge: Why do we need to change this to Array?
What will the command usage become with this change (could you please give an example)?
Also does it make sense to also update the tests and the docs if we make this change?

vladimir-kotikov: This just aligns `--archs` interpretation with `build` module (see [build.js#L156](https://github.com/apache/cordova-windows/blob/43b101a6cbd8392d61e51dca01f25296ae84c89b/template/cordova/lib/build.js#L156)). Also the logic below ([L65](https://github.com/apache/cordova-windows/pull/190/files#diff-31dc29ac3df976968b00e031ea8cde82R65)) translates `archs: String -> archs: String[]` so this change effectively won't change existing behavior of 'run' command

vladimir-kotikov: Update: in terms of nopt `String` argument type is the same as `[String]`. The latter syntax (with array literal) just allows to specify multiple allowed types

daserge: Makes sense :+1: LGTM

PR #191, opened 2016-08-04
CB-11658 activated event is not fired on Windows 10 RS1

daserge:
Patch start page to include base.js reference to HTML as a workaround

[Jira issue](https://issues.apache.org/jira/browse/CB-11658)


toddm:
:shipit:

PR #192, opened 2016-08-10
CB-11657 Add bom to www after plugin operations

vladimir-kotikov:
JIRA [CB-11657](https://issues.apache.org/jira/browse/CB-11657)

This PR addresses WACK validation issues when app has been built from VS after installing/restoring plugins. In this case some files, added by plugins would not have BOM


vladimir-kotikov:
@daserge. @sgrebnov, could you please take a look
/cc @toddm

daserge:
LGTM :+1:

toddm:
:shipit:

PR #195, opened 2016-09-08
CB-11836 Allow setting of 'ForegroundText' property via config.xml

tsschaffert:
We had a problem when trying to release an app in the Windows Store with BackgroundColor set to white because of the contrast between background and foreground color. With this PR, you can set the foreground color to 'dark' in these cases to pass store certification.

### DIFF
--- a/spec/unit/AppxManifest.spec.js
+++ b/spec/unit/AppxManifest.spec.js
@@ -187,6 +187,27 @@ describe('AppxManifest', function () {
         it('refineColor should leave CSS color name as is', function () {
             expect(refineColor(CSS_COLOR_NAME)).toEqual(CSS_COLOR_NAME);
         });
+        
+        it('setForegroundText should change the ForegroundText property on non-Windows 10 platforms', function () {
+            var visualElementsWindows = AppxManifest.get(WINDOWS_MANIFEST).getVisualElements();
+            var visualElementsWindows10 = AppxManifest.get(WINDOWS_10_MANIFEST).getVisualElements();
+            
+            var foregroundTextLight = 'light';
+            var foregroundTextDark = 'dark';
+            
+            // Set to 'light'
+            visualElementsWindows.setForegroundText(foregroundTextLight);
+            expect(visualElementsWindows.getForegroundText()).toEqual(foregroundTextLight);
+            
+            // Set to 'dark'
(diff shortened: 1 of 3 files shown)


daserge:
@tsschaffert, please let me know if you plan to address the comments.

tsschaffert:
@daserge I will look into it

tsschaffert:
1) The issue with Windows 10 should be fixed now, I handled it similarly to the 'ToastCapable' property.
2) The color argument is now passed to the manifest file unchecked, so invalid values return an error like 'App manifest validation failed.' when building the project. Alternatively, we could throw a TypeError for invalid values?
3) I added only a very basic test, as there were no comparable tests for the visualElement methods and I don't know enough about the internals to write really meaningful tests

template/cordova/lib/AppxManifest.js
```diff
@@ -390,11 +390,6 @@ AppxManifest.prototype.getVisualElements = function () {
         },
         setForegroundText: function (color) {
             if (color) {
```
daserge: Instead of `if (color) { visualElements.attrib.ForegroundText = color; }` would it be better to use `visualElements.attrib.ForegroundText = color || 'light'` so that ForegroundText would be reset as default `light` if we set `dark` via preference for example and then remove the preference from config.xml entirely?

spec/unit/AppxManifest.spec.js
```diff
@@ -187,6 +187,27 @@ describe('AppxManifest', function () {
...
+            var foregroundTextDark = 'dark';
+            
+            // Set to 'light'
+            visualElementsWindows.setForegroundText(foregroundTextLight);
+            expect(visualElementsWindows.getForegroundText()).toEqual(foregroundTextLight);
+            
+            // Set to 'dark'
+            visualElementsWindows.setForegroundText(foregroundTextDark);
+            expect(visualElementsWindows.getForegroundText()).toEqual(foregroundTextDark);
+
```
daserge: It also would be good to add a case for missing preference:
```javascript
            // Simulate the absence of preference, should default to 'light'
            visualElementsWindows.setForegroundText(undefined);
            expect(visualElementsWindows.getForegroundText()).toEqual(foregroundTextLight);
```

tsschaffert:
Good points, I updated the changes accordingly.

PR #196, opened 2016-09-14
Fix setSplashBgColor exception when no splashscreen is found

giuseppelt:
WindowsPhone allows to unset splashscreen (img) to avoid it to be shown.
This patch fixes whenever the "splashscreen" node is not present.

### DIFF
--- a/template/cordova/lib/AppxManifest.js
+++ b/template/cordova/lib/AppxManifest.js
@@ -391,10 +391,12 @@ AppxManifest.prototype.getVisualElements = function () {
         },
         setSplashBackgroundColor: function (color) {
             var splashNode = visualElements.find('./' + self.prefix + 'SplashScreen');
-            if (color && splashNode) {
-                splashNode.attrib.BackgroundColor = refineColor(color);
-            } else if (!color) {
-                delete splashNode.attrib.BackgroundColor;
+            if (splashNode) {
+                if (color) {
+                    splashNode.attrib.BackgroundColor = refineColor(color);
+                } else {
+                    delete splashNode.attrib.BackgroundColor;
+                }
             }
             return this;
         },


daserge:
Looks like SplashScreen node is required according to the [schema](https://msdn.microsoft.com/en-us/library/windows/apps/br211471.aspx).

giuseppelt:
Schema is incorrect or old.

According to https://msdn.microsoft.com/en-us/library/windows/apps/ff769511(v=vs.105).aspx splashscreen isn't required nor it's provided in the default template.

> Windows Phone 8 apps load quickly, so a Windows Phone 8 app usually doesn’t need a splash screen. New Windows Phone 8 projects don’t include a default splash screen image file. However, if you want to use a splash screen in your Windows Phone 8 app, add an existing JPG image file to the root folder of your Windows Phone 8 project, and name the file SplashScreenImage.jpg.

Anyway the patch here, doesn't delete the element, but just avoid null-reference when the node isn't present, like in the default template.

daserge:
That link is for Windows Phone 8 while cordova-windows is for Windows 8.1 / Windows Phone 8.1 and up.
In general the patch could be applicable for Windows 10 appxmanifest as SplashScreen is not required for it: https://msdn.microsoft.com/en-us/library/windows/apps/dn934817.aspx

Omitting it results in a black window frame color though which I don't see a way to avoid - and it seems like an issue to me.

giuseppelt:
This patch doesn't touch nothing, nor alter the manifest.
It's just a logic check, because the previous code assumes the node is always found.
The code simply is `if (node != null) ...dologic` nothing more than that.
The assumption is wrong, because you can have a manifest (WP8.1) without splashscreen node.

Manifest without the splashscreen node: --> Exception.
Manifest with the node: --> run as usual.

This patch is just a logic check, no omission, no something else.

vladimir-kotikov:
@giuseppelt, is there a JIRA issue about this? It would be nice if you could create one with explanation of your usecase and attach it to this PR by specifying JIRA item title at the beginning of PR name.

@daserge, could you please elaborate how this causes problems with window frame color?

daserge:
@giuseppelt thanks for the contribution - the logic fix is valid, I was not disputing this.
This requires additional changes though because the [current windows template](https://github.com/apache/cordova-windows/blob/258a21dbf6ac4bd020e304de8544992e5e3cd9f2/template/package.phone.appxmanifest#L50) includes SplashScreen node and giving the platforms are considered build artifacts - the node removal should be available through the config.xml preference.
That preference would be then documented in the cordova-plugin-splashscreen along with the [Windows 10 quirk](https://github.com/apache/cordova-windows/pull/196#issuecomment-254519722).

daserge:
no-splashscreen-node-win10

giuseppelt:
Jira issue https://issues.apache.org/jira/browse/CB-12036 set as Minor.

But, please, don't delay because the missing feature. The issue is beyond that, the code is wrong because the assumption is false.

vladimir-kotikov:
Merged, thanks!

PR #202, opened 2016-10-17
CB-11933: Remove capabilities from manifest

matrosov-nikita:
### What does this PR do?
Windows has special logic for appxmanifest's capabilities removal, therefore we should override this behaviour

Auto test

### Checklist


daserge:
Thanks, @matrosov-nikita !
I've tested and this works in terms of removing the added Capabilities.
Although the issue described in https://issues.apache.org/jira/browse/CB-11933 remains - that uap: prefixes are not added on plugin add.
So if you create a project in CLI, add a plugin and then open solution in Visual Studio - capabilities will not actually work.

vladimir-kotikov:
LGTM

vladimir-kotikov:
Decided to merge this as-is to at least have the fix for capabilities removal.

PR #212, opened 2016-11-25
CB-9287 Not enough Icons and Splashscreens for Windows 8.1 and Windows Phone 8.1

daserge:
### Platforms affected
Windows

### What does this PR do?
Add JPG support, added missing splashscreen scales for projects
cordova.js - refine splashscreen extension based on config.xml
Added a warning when non-MRT splash screens have mixed extensions
Added a warning for size limit exceeding (applies to win10 project only)
Added a warning for images with unsupported format
Tests

[Jira issue CB-9287](https://issues.apache.org/jira/browse/CB-9287)

Manual and auto tests

### Checklist


template/cordova/lib/prepare.js
```diff
@@ -439,6 +490,109 @@ function getUAPVersions(config) {
...
+
+    var desktopSplashScreen = splashScreens.filter(function(img) { 
+        return checkTargetMatchAndUpdateUsedExtensions(img, SPLASH_SCREEN_DESKTOP_TARGET_NAME);
+    })[0];
+
+    var phoneSplashScreen = splashScreens.filter(function(img) { 
+        return checkTargetMatchAndUpdateUsedExtensions(img, SPLASH_SCREEN_PHONE_TARGET_NAME);
+    })[0];
+
+    checkThatExtensionsAreNotMixed();
```
vladimir-kotikov: Could you please explain, what's the purpose of this check? Please give a couple of examples when it would yield a warning?

template/cordova/lib/prepare.js
```diff
@@ -439,6 +490,109 @@ function getUAPVersions(config) {
...
+
+    var manifestSplashScreenMap = [{
```
vladimir-kotikov: Nit: IMO dictionaly is a better candidate here than array of dictionaries w/ 2 keys each

template/CordovaApp.Phone.jsproj
```diff
@@ -79,7 +79,9 @@
     <AppxManifest Include="package.phone.appxmanifest">
       <SubType>Designer</SubType>
     </AppxManifest>
-    <Content Include="images\*.png" Exclude="images\*.scale-180.*" />
+    <Content Exclude="images\*SplashScreen.*" Include="images\*.png" />
```
daserge: We are excluding images named like `SplashScreen.jpg` or `SplashScreen.scale-100.png` here for phone project (we can do this taking into account that phone and desktop projects have different hardcoded splashscreen names).

template/cordova/lib/prepare.js
daserge:
```xml
<splash src="res/screen/windows/splashscreen.png" width="620" height="300"/>
<splash src="res/screen/windows/splashscreen.scale-125.jpg" width="775" height="375"/>
```

You should not mix image extensions for non-MRT syntax inside one project group - in this case ^ the second JPG image will not be used as AppxManifest will have PNG for SplashScreen.Image' extension.
There is also a [test for this case](https://github.com/apache/cordova-windows/pull/212/files#diff-954e5996601fc5ce41f4107af4f31525R620).

template/cordova/lib/prepare.js
```diff
@@ -353,6 +391,13 @@ function mapImageResources(images, imagesDir) {
                 events.emit('warn', 'No images found for target: ' + img.target);
             } else {
                 candidates.forEach(function(mrtImage) {
+                    if (img.target === SPLASH_SCREEN_DESKTOP_TARGET_NAME &&
+                        mrtSplashScreenToTargetProject(mrtImage) === TARGET_PROJECT_10 &&
+                        exceedsSizeLimit(mrtImage.path)) {
```
daserge: This does not result in a build failure and the issue could only be seen in Visual Studio manifest designer so I believe the warning is useful here.
Do you mean we should fail the build rather than showing this warning?
It currently just skips oversized images.

daserge:
@vladimir-kotikov thank you for the review!
I've addressed the notes.

template/CordovaApp.Phone.jsproj
vladimir-kotikov: Oh, I see. I didn't notice first time that you're excluding splashscreens for desktop app

daserge:
@vladimir-kotikov could you please take a look again?
I've also added a fix for [CB-12192 - No SplashScreen on Windows when content.src is subpage](https://issues.apache.org/jira/browse/CB-12192) - the issue occured while testing splashscreen API.

template/cordova/lib/prepare.js
vladimir-kotikov: > You should not mix image extensions for non-MRT syntax inside one project group

I thought it's pretty obvious :)

> in this case ^ the second JPG image will not be used as AppxManifest will have PNG for SplashScreen.Image' extension

Yep. If you specify the same splascreens using MRT and 'target' attribute, `splashscreen.scale-125.jpg` also would be silently ignored because `mapImageResources` requires extensions to match. That said I don't think we need to care about this at all.

daserge: In fact I've added this for a case when a developer could try to use JPG only for larger images (scale 200% and up) because they usually exceed the 200K limit.

template/cordova/lib/AppxManifest.js
```diff
@@ -416,6 +417,18 @@ AppxManifest.prototype.getVisualElements = function () {
             }
             return this;
         },
+        getSplashScreenExtension: function (extension) {
```
daserge: Please see https://github.com/apache/cordova-windows/pull/212#discussion_r90006200

daserge:
@vladimir-kotikov addressed, thanks! It is good to go?

vladimir-kotikov:
LGTM

frederikbosch:
Any chance that a version with this included will be released any time soon?

PR #213, opened 2016-12-06
CB-12163 Make resource-file copy files again

ktop:
### Platforms affected
- Windows

### What does this PR do?
- Move the current functionality of `resource-file` tag, which is referencing files, to be set by an attribute flag, and adding the copy functionality back in as the default behavior of resource-file. 

ex. 
This will default to copy: 
`<resource-file src="x64/foo.dll" target="x64/foo.dll" arch="x64" />`

And this will reference: 
`<resource-file src="x64/foo.dll" target="foo.dll" arch="x64" reference="true" />`

Will need to pull in the change from cordova-common for this to work: https://github.com/apache/cordova-lib/pull/509

- Updated the unit tests and ran them
- Created a new app with the updated resource-file and check to see if it copies files. 

### Checklist
- [X] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [X] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [X] Added automated test coverage as appropriate for this change.


ktop:
@vladimir-kotikov Can you review? This pr reverts resource-file to what it used to do so it will copy files.

daserge:
@ktop, could you please comment on @timbarham message?
http://markmail.org/message/2fproqgmbrc2gxpz
Is this a breaking change?

ktop:
@daserge Thanks for the bump, Tim's email was in my spam. Yes, this would be a breaking change as it is now. We should wait until we come to a final decision before proceeding with this pr.

template/cordova/lib/JsprojManager.js
```diff
@@ -121,7 +121,12 @@ jsprojManager.prototype = {
         copyToOutputDirectory.text = 'Always';
         children.push(copyToOutputDirectory);
 
-        var item = createItemGroupElement('ItemGroup/Content', sourcePath, targetConditions, children);
+        var item;
```
vladimir-kotikov: IMO it'd be better to calculate path for `content` element outside this method (in `'resource-file'.install`) - this way you will avoid checking the same condition (`if (targetConditions.reference) ...`) twice and won't need to modify this method at all.

template/cordova/lib/PluginHandler.js
```diff
@@ -53,18 +53,37 @@ var handlers = {
...
-            // do not copy, but reference the file in the plugin folder. This allows to
-            // have multiple source files map to the same target and select the appropriate
-            // one based on the current build settings, e.g. architecture.
-            // also, we don't check for existence. This allows to insert build variables
-            // into the source file name, e.g.
-            // <resource-file src="$(Platform)/My.dll" target="My.dll" />
-            var relativeSrcPath = getPluginFilePath(plugin, obj.src, project.projectFolder);
-            project.addResourceFileToProject(relativeSrcPath, obj.target, getTargetConditions(obj));
+            var targetConditions = getTargetConditions(obj);
+            if (targetConditions.reference) {
```
vladimir-kotikov: I think that the object that `getTargetConditions` returns was originally meant to hold details about target platform (Win8.1, WP8.1, Win10) and architecture to create a `condition` attribute, so `reference` field  doesn't really fit here - you might just check `if (obj.reference) ...`

template/cordova/lib/PluginHandler.js
```diff
@@ -53,18 +53,37 @@ var handlers = {
...
+                // into the source file name, e.g.
+                // <resource-file src="$(Platform)/My.dll" target="My.dll" />
+                var relativeSrcPath = getPluginFilePath(plugin, obj.src, project.projectFolder);
+                project.addResourceFileToProject(relativeSrcPath, obj.target, targetConditions);
+            } else {
+                // if target already exists, emit warning to consider using a reference instead of copying
+                if (fs.existsSync(path.resolve(project.root, obj.target))) {
+                    events.emit('warn', '<resource-file> with target ' + obj.target + ' already exists and will be overwritten ' +
+                    'by a <resource-file> with the same target. Consider using the attribute reference="true" in the ' +
+                    '<resource-file> tag to avoid overwriting files with the same target. ');
```
vladimir-kotikov: Might be also useful to add that adding `reference` will not copy user files to destination directory. WDYT?

template/cordova/lib/PluginHandler.js
```diff
@@ -188,7 +207,7 @@ module.exports.getUninstaller = function(type) {
 };
 
 function getTargetConditions(obj) {
-    return { versions: obj.versions, deviceTarget: obj.deviceTarget, arch: obj.arch };
+    return { versions: obj.versions, deviceTarget: obj.deviceTarget, arch: obj.arch, reference: obj.reference };
```
vladimir-kotikov: As per comments above `reference` field is not really required - i's easier to use `obj.reference` directly

vladimir-kotikov:
Looks good 👍

daserge:
@ktop what is the status of this PR? Is it ready for merge?

ktop:
@daserge we need to wait for a cordova-common release since the commit I'm dependent on has been merged https://github.com/apache/cordova-lib/pull/509

ktop:
@daserge 
Now that there has been a cordova-common release, can we refresh cordova-common in cordova-windows and get this merged?

vladimir-kotikov:
@ktop, done!

PR #244, opened 2017-12-07
CB-13641: (windows) allow for transparent splash screen background co…

IndieSW:
…lor.

### Platforms affected
windows

### What does this PR do?
allows for configuration of transparent splash screen background color.

built app with forked/updated repository and manually tested splash screen background color configuration with RGB and 'transparent' values.

### Checklist
- [x ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ x] Added automated test coverage as appropriate for this change.

### DIFF
--- a/cordova-js-src/splashscreen.js
+++ b/cordova-js-src/splashscreen.js
@@ -69,18 +69,21 @@ function readPreferencesFromCfg(cfg, manifest) {
         splashImageSrc = schema + ':///' + manifest.getSplashScreenImagePath().replace(/\\/g, '/');
 
         bgColor = cfg.getPreferenceValue('SplashScreenBackgroundColor') || bgColor;
-        bgColor = bgColor.replace('0x', '#').replace('0X', '#');
-        if (bgColor.length > 7) {
-            // Remove aplha
-            bgColor = bgColor.slice(0, 1) + bgColor.slice(3, bgColor.length);
-        }
+        bgColor = bgColor.toLowerCase().replace('0x', '#');
 
-        titleBgColor = {
-            a: 255,
-            r: parseInt(bgColor.slice(1, 3), 16),
-            g: parseInt(bgColor.slice(3, 5), 16),
-            b: parseInt(bgColor.slice(5, 7), 16)
-        };
+        if (bgColor !== 'transparent') {
+            if (bgColor.length > 7) {
+                // Remove aplha
+                bgColor = bgColor.slice(0, 1) + bgColor.slice(3, bgColor.length);
+            }
+
+            titleBgColor = {
+                a: 255,
+                r: parseInt(bgColor.slice(1, 3), 16),
+                g: parseInt(bgColor.slice(3, 5), 16),
+                b: parseInt(bgColor.slice(5, 7), 16)
+            };
+        }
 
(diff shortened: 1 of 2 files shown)


cordova-js-src/splashscreen.js
```diff
@@ -69,18 +69,21 @@ function readPreferencesFromCfg(cfg, manifest) {
         splashImageSrc = schema + ':///' + manifest.getSplashScreenImagePath().replace(/\\/g, '/');
 
         bgColor = cfg.getPreferenceValue('SplashScreenBackgroundColor') || bgColor;
-        bgColor = bgColor.replace('0x', '#').replace('0X', '#');
-        if (bgColor.length > 7) {
-            // Remove aplha
-            bgColor = bgColor.slice(0, 1) + bgColor.slice(3, bgColor.length);
-        }
+        bgColor = bgColor.toLowerCase().replace('0x', '#');
```
janpio: This unrelated change doesn't add the possibility of other bugs?

cordova-js-src/splashscreen.js
```diff
@@ -69,18 +69,21 @@ function readPreferencesFromCfg(cfg, manifest) {
...
+
+        if (bgColor !== 'transparent') {
+            if (bgColor.length > 7) {
+                // Remove aplha
```
janpio: maybe fix that typo while you're at it ;)

cordova-js-src/splashscreen.js
```diff
@@ -212,7 +215,7 @@ function exitFullScreen() {
 // Make title bg color match splashscreen bg color
 function colorizeTitleBar() {
     var appView = Windows.UI.ViewManagement.ApplicationView.getForCurrentView();
-    if (isWin10UWP) {
+    if (isWin10UWP && (typeof titleBgColor !== 'undefined')) {
```
janpio: no nicer way to do this but then to check for undefined?
took some time to understand what is going on here, especially with the `titleInitialBgColor ` that depends on this one

cordova-js-src/splashscreen.js
IndieSW: I had to force the case of bgColor to compare it to 'transparent' so I combined it with the replacement of '0x' and '0X'. bgColor is used as CSS color value which I believe is case-insensitive.

cordova-js-src/splashscreen.js
IndieSW: :)

cordova-js-src/splashscreen.js
IndieSW: That's a reasonable point. Both `titleBgColor` and `titleInitialBgColor` are set in one place and used in one place. Adding a variable to track whether or not `bgColor` is transparent seemed like too much. Thoughts?

janpio: First I thought you could just check if bgColor == 'transparent`, but that doesn't exist here. `isTitleBgTransparent` would be a solution and probably more readable, also makes the special case explicit.

IndieSW:
Closing for update.

janpio:
Hm? You could have just committed the changes to your `IndieSW:master` branch for them to show up here automatically.

PR #239, opened 2017-08-11
CB-13175: fixing windows8.1 crash on startup

kaplanmaxe:
### Platforms affected

windows

### What does this PR do?

Ticket: https://issues.apache.org/jira/browse/CB-13175
 
This issue is a follow-up to [CB-12784](https://github.com/apache/cordova-windows/pull/232)

The user added the fix in `cordova-js-src/splashscreen.js` however it needs to be added to `template/www/cordova.js` as well. This PR takes the already accepted fix and applies it to `cordova.js` as well.

### What testing has been done on this change?

Ran on Windows 10 Creators Update (build 15063.250), Ran on Windows 8.1, Ran on Windows 10 mobile

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #246, opened 2018-01-25
CB-13829 Fix tests that were broken since CB-13237

janpio:
### Platforms affected
Windows 10, Visual Studio 2015 and 2017

### What does this PR do?
This PR fixes the e2e tests that were broken since https://issues.apache.org/jira/browse/CB-13237 was merged which was not caught because of broken CI.

For that I rewrote all of them from "default 8.1" to "default 10" by fixing the `expect`s, extracting the comparison stuff in its own function along the way. I think I replicated all test cases that were there before.

Local runs of e2e test suite + CI.

Note: The remaining 1 failing test with Visual Studio 2017 is from another, unrelated change (that was also not caught because of broken CI). I changed the order of tests so it is the last one being executed, so all the others are passing. This one will be tackled in https://issues.apache.org/jira/browse/CB-13834

### Checklist

### DIFF
--- a/spec/e2e/endtoend.spec.js
+++ b/spec/e2e/endtoend.spec.js
@@ -73,29 +73,29 @@ describe('Cordova create and build', function () {
 
     // default
 
-    it('spec.2 should build default (win10) project', function () {
+    it('spec.2a should build default (win10) project', function () {
         shell.exec(buildScriptPath + '', {silent: silent});
         _expectSubdirAndFileExist('CordovaApp.Windows10_1.0.0.0_anycpu_debug_Test', 'CordovaApp.Windows10_1.0.0.0_anycpu_debug.appx');
     });
 
     // --appx
 
-    it('spec.2a should build uap/win10 project', function () {
+    it('spec.2b should build uap/win10 project', function () {
         shell.exec(buildScriptPath + ' --appx=uap', {silent: silent});
         _expectSubdirAndFileExist('CordovaApp.Windows10_1.0.0.0_anycpu_debug_Test', 'CordovaApp.Windows10_1.0.0.0_anycpu_debug.appx');
     });
 
-    it('spec.2b should build 8.1 win project', function () {
+    it('spec.2c should build 8.1 win project', function () {
         shell.exec(buildScriptPath + ' --appx=8.1-win', {silent: silent});
         _expectExist(/.*Windows.*\.appxupload/);
     });
 
-    it('spec.2c should build 8.1 phone project', function () {
+    it('spec.2d should build 8.1 phone project', function () {
         shell.exec(buildScriptPath + ' --appx=8.1-phone', {silent: silent});
         _expectExist(/.*Phone.*\.appxupload/);
     });
 
-    it('spec.2c should build 8.1 win + phone project', function () {
+    it('spec.2e should build 8.1 win + phone project', function () {
         shell.exec(buildScriptPath + ' --appx=8.1', {silent: silent});
         _expectExist(/.*Windows.*\.appxupload/);
         _expectExist(/.*Phone.*\.appxupload/);
@@ -229,6 +229,7 @@ describe('Cordova create and build', function () {
         _expectSubdirAndFileExist('CordovaApp.Phone_1.0.0.0_x86_Test', 'CordovaApp.Phone_1.0.0.0_x86.appx');
     });
 
+    // this will be move up again when it is fixed for VS 2017
     it('spec.6a should generate appxupload and appxbundle for Windows 10 project bundle release build', function () {
         shell.exec(buildScriptPath + ' --release --bundle --archs=\"x64 x86 arm\"', {silent: silent});
         _expectExist(/.*bundle\.appxupload$/, 3);
--- a/spec/e2e/endtoend.spec.js
+++ b/spec/e2e/endtoend.spec.js
@@ -79,7 +79,7 @@ describe('Cordova create and build', function () {
 
     // --appx
 
-    it('spec.2a should build win10 project', function () {
+    it('spec.2a should build uap/win10 project', function () {
         shell.exec(buildScriptPath + ' --appx=uap', {silent: silent});
         _expectSubdirAndFileExist('CordovaApp.Windows10_1.0.0.0_anycpu_debug_Test', 'CordovaApp.Windows10_1.0.0.0_anycpu_debug.appx');
     });
@@ -107,18 +107,18 @@ describe('Cordova create and build', function () {
         _expectSubdirAndFileExist('CordovaApp.Windows10_1.0.0.0_x64_debug_Test', 'CordovaApp.Windows10_1.0.0.0_x64_debug.appx');
     });
 
-    it('spec.3b should build project for particular CPU', function () {
+    it('spec.3b should build project (8.1) for particular CPU', function () {
         shell.exec(buildScriptPath + ' --appx=8.1 --archs=\"x64\"', {silent: silent}); /* eslint no-useless-escape : 0 */
         _expectExist(/.*Phone.*x64.*\.appxupload/);
         _expectExist(/.*Windows.*x64.*\.appxupload/);
     });
 
-    it('spec.3c should build project for particular CPU', function () {
+    it('spec.3c should build project (8.1-win) for particular CPU', function () {
         shell.exec(buildScriptPath + ' --appx=8.1-win --archs=\"x64\"', {silent: silent}); /* eslint no-useless-escape : 0 */
         _expectExist(/.*Windows.*x64.*\.appxupload/);
     });
 
-    it('spec.3d should build project for particular CPU', function () {
+    it('spec.3d should build project (8.1-phone) for particular CPU', function () {
         shell.exec(buildScriptPath + ' --appx=8.1-phone --archs=\"x64\"', {silent: silent}); /* eslint no-useless-escape : 0 */
         _expectExist(/.*Phone.*x64.*\.appxupload/);
     });
@@ -131,7 +131,7 @@ describe('Cordova create and build', function () {
         _expectSubdirAndFileExist('CordovaApp.Windows10_1.0.0.0_anycpu_debug_Test', 'CordovaApp.Windows10_1.0.0.0_anycpu_debug.appx');
     });
 
-    it('spec.4b should build project for CPUs separated by whitespaces', function () {
+    it('spec.4b should build project (8.1) for CPUs separated by whitespaces', function () {
         shell.exec(buildScriptPath + ' --appx=8.1 --archs=\"x64 x86 arm anycpu\"', {silent: silent}); /* eslint no-useless-escape : 0 */
         _expectExist(/.*Phone.*x86.*\.appxupload/);
         _expectExist(/.*Phone.*x64.*\.appxupload/);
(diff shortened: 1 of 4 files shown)


brodybits:
@janpio can you give any explanation of how reverting dd91933 ("CB-13799 Updated checked-in node_modules") in 18eb64389700667608c0186da1e5112afc67e23e could trigger the test failures?

Also can you also give me an idea what dd91933 was supposed to do in the first place?

janpio:
https://github.com/apache/cordova-windows/commit/dd91933d7030f6f0bb207a676968c4fdad9e3691 has nothing to do with this, it was just a commit of the release process that is documented here: https://github.com/apache/cordova-coho/blob/master/docs/platforms-release-process.md#update-and-pin-dependencies It had to be reverted as the release process was canceled as I discovered the failing tests and breaking changes (which requires a 6.0 release instead of a 5.1 I was working on).

The culprit for the failing tests is in https://issues.apache.org/jira/browse/CB-13237 / https://github.com/apache/cordova-windows/commit/c1dfc77ddf866b4872330eb2b8874ede05daddcd but wasn't noticed as earlier `jasmine-node` was replaced by `jasmine` which doesn't have the same command line params, which caused the tests to fail silently and not to be run at all via https://github.com/apache/cordova-windows/commit/1c6d245fbb10ab71c77eed6a73989827042422f5

brodybits:
Thanks @janpio. Can you give me an idea of what portion of tests are now passing vs failing?

janpio:
Look below - there is a "Details" link next to the checks that go to the actual test runs. All but one are passing, but this will be covered in https://issues.apache.org/jira/browse/CB-13834 - this PR is about the changes to the tests to match the actual implementation.

brodybits:
All but one are passing, but this will be covered in
https://issues.apache.org/jira/browse/CB-13834 - this PR is about the
changes to the tests to match the actual implementation.


Thanks I will review later today. It would be ideal if you can rebase on
the latest changes on master.

janpio:
Done (although there are no relevant changes that would require this in that last commit).

janpio:
Done (although there are no relevant changes that would require this in that last commit).

spec/e2e/endtoend.spec.js
```diff
@@ -54,34 +71,82 @@ describe('Cordova create and build', function () {
...
     });
 
-    it('spec.2 should build project', function () {
-        shell.exec(buildScriptPath, {silent: true});
-        var packages = shell.ls(appPackagesFolder);
-        expect(packages.filter(function (file) { return file.match(/.*Phone.*\.appx.*/); }).length).toBe(1);
-        expect(packages.filter(function (file) { return file.match(/.*Windows.*\.appx.*/); }).length).toBe(1);
+    // default
+
+    it('spec.2 should build default (win10) project', function () {
```
brodybits: I wonder if this should already be spec.2a?

spec/e2e/endtoend.spec.js
```diff
@@ -54,34 +71,82 @@ describe('Cordova create and build', function () {
...
+        _expectExist(/.*Windows.*\.appxupload/);
+    });
+
+    it('spec.2c should build 8.1 phone project', function () {
+        shell.exec(buildScriptPath + ' --appx=8.1-phone', {silent: silent});
+        _expectExist(/.*Phone.*\.appxupload/);
     });
 
-    it('spec.5 should build project containing plugin with InProcessServer extension', function (done) {
+    it('spec.2c should build 8.1 win + phone project', function () {
```
brodybits: spec.2c repeated

spec/e2e/endtoend.spec.js
janpio: doesn't make a difference really, but I think all the others start with `a` as well, so might as well do this here

spec/e2e/endtoend.spec.js
```diff
@@ -34,15 +34,32 @@ describe('Cordova create and build', function () {
     var buildDirectory = path.join(__dirname, '../..');
     var appPackagesFolder = path.join(buildDirectory, projectFolder, 'AppPackages');
     var buildScriptPath = '"' + path.join(buildDirectory, projectFolder, 'cordova', 'build') + '"';
+    var silent = false;
+
+    function verifySubDirContainsFile (subDirName, fileName, count) {
+        count = typeof count !== 'undefined' ? count : 1;
```
brodybits: ugly overwrite (I think)

I would propose something like this:

```js
var expectedCount = typeof count !== 'undefined' ? count : 1;
```

spec/e2e/endtoend.spec.js
```diff
@@ -34,15 +34,32 @@ describe('Cordova create and build', function () {
...
 
-    function verifySubDirContainsFile (subDirName, fileName) {
         var subDir = path.join(appPackagesFolder, subDirName);
         var packages = shell.ls(subDir);
-        expect(packages.filter(function (file) { return file.match(fileName); }).length).toBe(1);
+        expect(packages.filter(function (file) { return file.match(fileName); }).length).toBe(count);
+    }
+
+    function _expectExist (fileNamePattern, count) {
+        count = typeof count !== 'undefined' ? count : 1;
```
brodybits: ditto

spec/e2e/endtoend.spec.js
janpio: that line is actually just an ugly "hack" to get default params for older node versions, otherwise one could just `count = 1` in the param list of the functions. yay JavaScript.

spec/e2e/endtoend.spec.js
```diff
@@ -54,34 +71,82 @@ describe('Cordova create and build', function () {
...
+    it('spec.2c should build 8.1 win + phone project', function () {
+        shell.exec(buildScriptPath + ' --appx=8.1', {silent: silent});
+        _expectExist(/.*Windows.*\.appxupload/);
+        _expectExist(/.*Phone.*\.appxupload/);
+    });
+
+    // --archs
+
+    it('spec.3a should build project for particular CPU', function () {
+        shell.exec(buildScriptPath + ' --archs=\"x64\"', {silent: silent}); /* eslint no-useless-escape : 0 */
```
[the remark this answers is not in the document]
janpio: Re eslint: No idea, was there already.
Re extract into function: Out of scope of this issue, but good idea for follow up PR.

spec/e2e/endtoend.spec.js
```diff
@@ -54,34 +71,82 @@ describe('Cordova create and build', function () {
...
+    });
+
+    it('spec.3d should build project (8.1-phone) for particular CPU', function () {
+        shell.exec(buildScriptPath + ' --appx=8.1-phone --archs=\"x64\"', {silent: silent}); /* eslint no-useless-escape : 0 */
+        _expectExist(/.*Phone.*x64.*\.appxupload/);
+    });
+
+    it('spec.4a should build project for CPUs separated by whitespaces', function () {
+        shell.exec(buildScriptPath + ' --archs=\"x64 x86 arm anycpu\"', {silent: silent}); /* eslint no-useless-escape : 0 */
+        _expectSubdirAndFileExist('CordovaApp.Windows10_1.0.0.0_x64_debug_Test', 'CordovaApp.Windows10_1.0.0.0_x64_debug.appx');
```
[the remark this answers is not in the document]
janpio: out of scope, but also good candidate for follow up PR.

template/cordova/lib/build.js
```diff
@@ -411,92 +437,6 @@ function clearIntermediatesAndGetPackage (bundleTerms, config, hasAnyCpu) {
     return pckage.getPackageFileInfo(finalFile);
 }
 
-// Can update buildConfig in the following ways:
-//  * Sets targetProject property, the project to launch when complete
-function getBuildTargets (buildConfig) {
```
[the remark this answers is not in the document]
janpio: I think I chose this one to remove because the other one was closer to what will be kept in use. 
With the export I technically agree, but this is quite a mess anyway and should probably be cleaned up in a future PR.

brodybits:
Thanks @janpio for the nice work so far. I gave my review comments what I think are all minor (non-blocking), leaving it up to your best judgment. I am also looking forward to the fix for <https://issues.apache.org/jira/browse/CB-13834>. I am really hoping to see a release to fix the crash on Windows 8.1 (https://issues.apache.org/jira/browse/CB-13175 - fixed in PR #239).

janpio:
> I would favor making these changes in two parts (maybe split into 2 different issues): one to fix the actual Windows build (with just enough test coverage to prove it working) and another to fix and extend the test coverage.

In my opinion these changes actually are only to fix the previously existing tests, I just went through those that didn't work and replace them with the new equivalents for the new "UWP = standard" situation. I didn't really extend the test coverage, did I? If so, it was more an accident ;)

> I do not really like the idea of moving the failing test function if we expect to to move it right back. I would favor just marking it as pending (use xit or pending function) and then fixing it in issues.apache.org/jira/browse/CB-13834.

Yeah, in principle I agree - but I wanted to make sure the tests keep failing as that test is actually broken. So moving it down just enabled them all to run but it is still clearly visible that `master` is in a broken state.

janpio:
Thanks for the review @brodybits, very helpful.

I did some smaller changes, and think I will "ignore" the other things for now.

Do you think the "new"/rewritten tests are doing their job equivalent to what was there before? If you agree, I think I will just merge this in to see some progress here.

If you have any ideas about https://issues.apache.org/jira/browse/CB-13834... I am pretty much out of ideas as I can't reproduce locally and don't know what is going on on AppVeyor :/

brodybits:
> In my opinion these changes actually are only to fix the previously existing tests

It is clear to me that a major goal is to pass most of the existing test cases. It is not clear to me whether or not the following items are other goals of this change:
- fix actual script behavior
- fix test cases that were incorrect

> I didn't really extend the test coverage, did I?

It looks like the following test cases were added: 2b, 2c, 2d, 2e, 3b, 3c, 3d, 4b, 5b, 6b

It appears that the test coverage would be expanded to cover Windows 8.1, Windows Phone 8.1, Windows 10 and all possible CPU targets more completely.

> I did some smaller changes, and think I will "ignore" the other things for now.

Fine for me, they are all minor anyway.

> Do you think the "new"/rewritten tests are doing their job equivalent to what was there before?

Yes, with more complete coverage of Windows 8.1/Windows Phone 8.1/Windows 10 build and all CPU targets.

> If you agree, I think I will just merge this in to see some progress here.

Yes, please.

In general I think this is a major step in the right direction. I would be happy to take a look at CB-13834, cannot promise anything though. I am still hoping we can have a release within the next few days to solve the Windows 8.1 crash problem.

janpio:
> It looks like the following test cases were added: 2b, 2c, 2d, 2e, 3b, 3c, 3d, 4b, 5b, 6b

> It appears that the test coverage would be expanded to cover Windows 8.1, Windows Phone 8.1, Windows 10 and all possible CPU targets more completely.

Before the default was Win8.1, now it is Win10. That meant that the Win8.1 stuff that was automatically done before now has to be done explicitly. Also some of the architecture splits that were possible with default Win8.1 are not possible with Win10, so this had to be transformed as well.

> I am still hoping we can have a release within the next few days to solve the Windows 8.1 crash problem.

I can not do a release with failing tests which represent an actual problem, which CB-13834 seems to do. Either the test is broken or the actual implementation for VS2017 is broken - both won't allow me to cut a release.

The only way to do a 5.1 with the patch you need would be to go back to before the UWP (and probably also test) changes, then cherry pick the one you need and do a release only for with that. But that is definitely not a viable option though because of the work involved, sorry. So we nee to get CB-13834 fixed, then I can continue.

brodybits:
Thanks for all of your efforts so far, definitely appreciated! I would definitely agree that we need to deal with CB-13834 before releasing 5.1 and would be happy to take a look within the next week or so.

janpio:
(6.0 actually because the UWP change was a breaking one - but looking for any feedback on CB-13834)

PR #248, opened 2018-02-08
[WIP] CB-13834 TEST FOR DISCUSSION [DO NOT MERGE]

brodybits:
### Platforms affected

Windows

### What does this PR do?

Further testing to help narrow down the cause of CB-13834 (<https://issues.apache.org/jira/browse/CB-13834>) test failures

OBSERVATION: For some reason it would only generate x64 build for UWP bundle in case of VS 2017 on AppVeyor build

Output on my fork is available at: <https://ci.appveyor.com/project/brodybits/cordova-windows-u6yx9/build/1.0.13/job/vpstha4h1l4879hc> (same results are expected to show up for this PR)

NOTICE: This PR is for investigation and discussion only. PLEASE DO NOT MERGE.

- appveyor.yml replace Node 4 with Node 8 to rule out possible effect of outdated Node.js version
- additional logs (commented out) used to get a better understanding of actual output
- fixed related test descriptions
- added test of UWP bundle for x64 only (seems to pass on AppVeyor)
- add extra expectation to spec.6a to check for generated bundle for x64 only

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database (reported by someone else): <https://issues.apache.org/jira/browse/CB-13834>
~~Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.~~ not done here (raised for investigation and discussion only)


spec/e2e/endtoend.spec.js
```diff
@@ -203,7 +213,7 @@ describe('Cordova create and build', function () {
 
     // here be 6a
 
-    it('spec.6b should generate appxupload and appxbundle for Windows 10 project bundle release build', function () {
+    it('spec.6b should generate appxupload and appxbundle for Windows Phone 8.1 project bundle release build (all target platforms)', function () {
```
janpio: ah interesting, missed that before

janpio:
> OBSERVATION: For some reason it would only generate x64 build for UWP bundle in case of VS 2017 on AppVeyor build

Isn't this just because it is failing after building the first platform because of
`C:\Program Fil...ckage.Targets(2975,5): error MSB3231: Unable to remove directory "build\windows\bld\PackageUploadLayout\". The process cannot access the file 'Default.rd.xml' because it is being used by another process. [C:\pro...ndows10.jsproj]`?

brodybits:
If I would revert d5ddcb4 ([CB-12499](https://issues.apache.org/jira/browse/CB-12499)) then the failure reported in CB-13834 would go away.

So the question is why the file added by CB-12499 triggers the CI failure. (This is not a problem if I run the e2e test on my own system.) I have a hunch that this is something wrong with the VS 2017 installation on AppVeyor, would probably need a simpler test project to verify this. I think it comes down to trying one thing at a time until we know what does and does not work.

@janpio would you consider reverting CB-12499 for now to unblock the badly needed cordova-windows release?

[1] <https://ci.appveyor.com/project/brodybits/cordova-windows-u6yx9/build/1.0.14>

janpio:
This part of the discussion should not happen here bit in https://issues.apache.org/jira/browse/CB-13834. Please repost the relevant parts there, otherwise noone will be able to find this in the future (including myself).

brodybits:
> This part of the discussion should not happen here bit in https://issues.apache.org/jira/browse/CB-13834.

My comments are automatically reflected in <https://issues.apache.org/jira/browse/CB-13834>.

janpio:
Just looked over your PR and changes again:

> added test of UWP bundle for x64 only (seems to pass on AppVeyor)

Are these the 3 tests with `pre` that are run before spec6? 
Were these to check if the "architectures" can be built one by one, but not when combined into one "call"?

brodybits:
Were these to check if the "architectures" can be built one by one, but not
when combined into one "call"?


Yes

PR #284, opened 2018-07-27
CB-14224 Default.rd.xml header fixes

brodybits:
### Original issue

<https://issues.apache.org/jira/browse/CB-14224>

### Platforms affected

Windows

### What does this PR do?

`template/Properties/Default.rd.xml` header fixes:
- add MIT license text from <https://github.com/Microsoft/Windows-universal-samples>, with left aligment fixed
- add link to <https://github.com/Microsoft/Windows-universal-samples/blob/master/SharedContent/cs/Default.rd.xml>

Rationale: I think it is desired to add license headers whenever possible to satisfy Apache RAT license auditing tool. While I think it should be possible to configure the auditing tool to skip this file I would rather follow the practice of adding the license text whenever possible.

NOTE: I also proposed updates to LICENSE text in <https://github.com/Microsoft/Windows-universal-samples/pull/949> _(not included in this PR)_.

I verified that `coho audit-license-headers -r windows` with these changes does not show the missing license header any more.

### Checklist

- ~~[ ] Added automated test coverage as appropriate for this change.~~

### DIFF
--- a/template/Properties/Default.rd.xml
+++ b/template/Properties/Default.rd.xml
@@ -13,6 +13,28 @@
 
     Using the Namespace directive to apply reflection policy to all the types in a particular namespace
     <Namespace Name="DataClasses.ViewModels" Seralize="All" />
+
+    The MIT License (MIT)
+
+    Copyright (c) Microsoft Corporation
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


brodybits:
Thanks for the quick review. I am afraid that I don't really understand
your question. I think this piece of XML was intended as some kind of
optional instructions.

brodybits:
As this change is already part of 6.0.x (GH-285) it should now be in master. Title and description have been updated.

PR #293, opened 2018-09-11
Remove bundled dependencies

raphinesse:
Part of https://github.com/apache/cordova/issues/6.

The only dependency versions that differed from the bundled ones after a normal installation at the time of writing were `cordova-common` which changed from 2.2.1 to 2.2.5 and `semver` which changed from 5.5.0 to 5.5.1.


janpio:
I fear this is responsible for https://github.com/apache/cordova-windows/issues/326 :(

PR #318, opened 2018-12-19
[WIP] Try out Travis windows support

purplecabbage:
### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+os: windows
 sudo: false
 
 git:


timbru31:
I've just restarted the Travis builds and they succeeded! (at least the 
continuous-integration/travis-ci/pr) 🎉 - I would assume that these issues/failrues from before could be TravisCI issues since back in December the Windows support was very, very new and in nearly half a year things got more stable.

timbru31:
Okay revert my first comment. :(

Why does https://travis-ci.org/apache/cordova-windows/builds/469915419 fail but https://travis-ci.org/apache/cordova-windows/builds/469915465 succeed? Where is the difference between `/travis-ci/push` and `/travis-ci/pr`?

janpio:
Rebasing the branch onto `master` should bring those two builds in sync with the same result.

timbru31:
Yes, but I’ve had not enough time earlier. 😊
IMO this is good to be merged

janpio:
Merging this doesn't make to much sense, as it just changes the environment from Linux to Windows. It would be better if we could _add_ Windows as another environment for Travis (although we could of course first discuss if this actually makes sense) and more importantly actually test something, similar to what happens on AppVeyor.

timbru31:
But running the tests on windows makes much more sense than testing that the tests for windows pass on Linux?
Of course we could expand the matrix of Travis to test both OS

Tbh I haven’t looked at the AppVeyor config if there are additional tests

purplecabbage:
This pr was never intended to be merged, it was to test windows support in travis.
Linux testing doesn't make sense because it just skips over any real tests, and just does linting.
Realistically .travis.yml should look like appveyor.yml on os:windows

janpio:
> But running the tests on windows makes much more sense than testing that the tests for windows pass on Linux?

The tests on Travis are different from the AppVeyor one - they are just testing the platform independent stuff (API etc). On AppVeyor it actually builds apps.

> Of course we could expand the matrix of Travis to test both OS

Looking forward to that PR - that would indeed be worth just to get an alternative to AppVeyor.

> Tbh I haven’t looked at the AppVeyor config if there are additional tests

Looking at the build output should give you an idea. Having this on Travis CI would give use more environment variations which would be very valuable - because Cordova Windows tends to be brittle there (see current CI failures on master).

---

Thanks @purplecabbage for kicking this off anyway.

PR #330, opened 2019-04-26
Conditionally skip the _one_ broken test for VS 2017 instead of allowing the whole suite to fail

raphinesse:
### Motivation and Context

The CI tests on AppVeyor using VS 2017 were allowed to fail completely because of _one_ E2E test that is broken under these conditions. This is confusing and could mask any other broken tests.

Example: https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-windows/builds/23755792

### Description

Instead of allowing the whole suite to fail when using VS 2017, we only skip the known-bad tests when VS 2017 on AppVeyor is used.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,10 +31,6 @@ environment:
     - nodejs_version: "6"
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
 
-matrix:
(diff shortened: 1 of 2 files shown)


janpio:
TIL `pending()` is a thing. Nice!

PR #254, opened 2018-02-13
CB-13875 add `prepare` script that can be called in e2e tests

janpio:
… to prepare project files

(based on `clean` and some terrible code to build a `project` object)
also adds default `config.xml` as in other projects

replaces #250

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

Issue #334, opened 2019-05-06
AppVeyor CI is failing because of newer MSBuild getting selected


janpio:
CI started failing when https://github.com/apache/cordova-windows/pull/329 was merged:

https://github.com/apache/cordova-windows/commit/95401b36b20e5c4b37a3253ed0042b98380caf27
https://ci.appveyor.com/project/Humbedooh/cordova-windows/builds/24305931

Example:
```
.Creating Cordova Windows Project:
	Path: temp\platforms\testcreate 応用
	Namespace: com.test.app
	Name: 応用
Windows project created with cordova-windows@7.1.0-dev
Building project: C:\projects\cordova-windows\temp\platforms\testcreate 応用\CordovaApp.Windows.jsproj
	Configuration : debug
	Platform      : anycpu
	Buildflags    : /p:AppxBundle=Never
	MSBuildTools  : C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin
buildProject spawn: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\msbuild [ 'C:\\projects\\cordova-windows\\temp\\platforms\\testcreate 応用\\CordovaApp.Windows.jsproj',
  '/clp:NoSummary;NoItemAndPropertyList;Verbosity=minimal',
  '/nologo',
  '/p:Configuration=debug',
  '/p:Platform=anycpu',
  '/p:AppxBundle=Never' ] { stdio: 'inherit' }
  prebuild.js: Patching platform `8.1`
  - Injected `base.js` reference to `www/index.html`
  - Removing /( *)(<script\s+(?:type="text\/javascript"\s+)?src="\/www\/WinJS\/js\/base.js">\s*<\/script>)(\s*)/ from www/index.html
  - Removing /( *)(<script\s+(?:type="text\/javascript"\s+)?src="\/\/Microsoft.Phone.WinJS.2.1\/js\/base.js">\s*<\/script>)(\s*)/ from www/index.html
[...]

PR #335, opened 2019-05-07
Fix wrong (default) description content being written to project files

### LINKED ISSUE #270: description tag set to default instead of defined value - regression from 5.0.0
The `<description>` tag is set to a default value instead of the in the config.xml defined value. This is is a regression from cordova-windows 5.0.0.

### situation
macOS host machine
cordova-lib: 8.1.0
cordova-windows 6.0.0

A description is correctly set in config.xml:

config.xml
```
  <description>
    Demo
  </description>
```

when running on a mac (or windows):
```
cordova prepare
```

### actual

platforms/windows/config.xml contains an additional description tag (with default content), in addition to the customized description tag
```
  <description>
    A sample Apache Cordova application that responds to the deviceready event.
  </description>
  [...]
  <description>
    Demo
  </description>
```

The appxmanifest files contain only the default value instead of the set custom value:
`platforms/windows/package.phone.appxmanifest`
`platforms/windows/package.windows.appxmanifest`
`platforms/windows/package.windows10.appxmanifest`
```
-        <Description>Demo</Description>
+        <Description>A sample Apache Cordova application that responds to the deviceready event.</Description>
```
```
-            <m2:VisualElements Description="Demo" DisplayName="Demo">
+            <m2:VisualElements Description="A sample Apache Cordova application that responds to the deviceready event." DisplayName="Demo">
```
### expected

the correctly set description tag gets propagated.

### conditions

* happens only with cordova-windows 6.0.0 - doesn't happen when using cordova-windows 5.0.0 with an otherwise unchanged setup

* happens on mac & windows - however this a serious problem, as the generated code could be committed by mistake

* probable cause: in 6.0.0 the `platforms/windows/cordova/defaults.xml` contains a default value for the desciption tag for the first time.

The wrong description value comes form the defaults.xml, however the `<name>` and `<author>` tags get processed correctly:
`platforms/windows/cordova/defaults.xml`
```
    <description>
        A sample Apache Cordova application that responds to the deviceready event.
    </description>
```

janpio:
### Motivation and Context

This PR fixes a bug that led to the wrong app description being used for Cordova Windows apps.

The bug was introduced in https://github.com/apache/cordova-windows/pull/254/files#diff-83c0809a5a3967aec365d13bf70b2be3. That change allowed the prepare script to work properly in tests, but broke the logic from the `config.xml` creation logic as I didn't also create a `defaults.xml` file at the same time that was empty and that otherweise would have been used as the initial template to build a `config.xml` from.

### Description

So this PR creates that missing `defaults.xml` file.

closes #270

### Testing

I tested this locally by using the platform to create a project and observed that it now uses the correct description in the Cordova Windows project files.

### Checklist

- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))


janpio:
This PR is failing as bad as current master, but that won't change before https://github.com/apache/cordova-windows/issues/334 is not implemented.

purplecabbage:
lgtm!

PR #264, opened 2018-02-16
[WIP] Win10 only

janpio:
### What does this PR do?

Removes all logic for non-Win10.

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #355, opened 2019-07-12
[WIP] README.md updates

brodybits:
- update requirements
- update link to Windows platform guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,17 +33,9 @@ This repo contains the code for an [Apache Cordova](http://cordova.apache.org) p
 
 # Requirements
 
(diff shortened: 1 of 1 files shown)


brodybits:
This is now WIP since the test suite seems to cover some usage on Windows 8.1, which I thought we should have removed. I thought we had dropped Windows 8.1 in the past. @janpio can you explain?

janpio:
Nobody dropped Windows 8.1 support - it is scatter over the whole codebase and just removing the conditionals makes code almost impossible to understand. (You are probably thinking of https://github.com/apache/cordova-windows/pull/264 which gave me that insight)

PR #358, opened 2019-07-15
fix(win10): read splashscreen correctly from manifest

### LINKED ISSUE #322: SplashScreen tag name does not match entry in Windows 10 Manifest file on some devices
The Windows 10 Manifest uses the _uap_ namespace for UAP/UWP tags. The following tag name for UWP in _confighelper.js_ does not match the _SplashScreen_ tag in that manifest, a modified _uap:SplashScreen_ tag name does.

https://github.com/apache/cordova-windows/blob/8cc627b8f5fb4a0959c4ef7a5994a4f358215fd6/template/www/cordova.js#L848

https://github.com/apache/cordova-windows/blob/8cc627b8f5fb4a0959c4ef7a5994a4f358215fd6/cordova-js-src/confighelper.js#L28

Results in _undefined_ access when retrieving splash screen image path etc. which causes a bunch of other issues ...

Don't know what the reason behind the differing behavior is, maybe the Windows 10 build version. Version 10.0.17736 requires the _uap_ namespace, version 10.0.17134 is only ok without the namespace, same app build.

janpio:
Currently it tries to read `SplashScreen` from the `.appxmanifest` file which does not exist, ut `uap:SplashScreen` does. This leads to an error, that has all kinds of follow up errors regarding the splashscreen functionality.

closes #322


brodybits:
Any pointers on how one of us can test this fix?

janpio:
Current release or master produces output like this: https://github.com/apache/cordova-windows/issues/322#issuecomment-511400869
Using this branch this will not happen any more.

brodybits:
> Current release or master produces output like this: [#322 (comment)](https://github.com/apache/cordova-windows/issues/322#issuecomment-511400869) (see screenshot)

Confirmed on my part

> Using this branch this will not happen any more.

Yes but I had to manually apply the patch to `platforms\windows\www\cordova.js` for that error message to go away. (Backslashes come from Windows cmder as I do not generally use Powershell or any form of Windows bash.)

This implies that whoever does the release has to use cordova-coho to regenerate [`template/www/cordova.js`](https://github.com/apache/cordova-windows/blob/master/template/www/cordova.js) (and should test it) to get this fixed in the next release.

janpio:
No, as https://github.com/apache/cordova-windows/issues/322#issuecomment-511400869 is still waiting for feedback from the initial reporter of the problem.

regnete:
Please merge this.

PR #360, opened 2019-07-15
Requirements link in README.md

brodybits:
UPDATED: replace existing requirements with link, according to the current situation


brodybits:
Changed to just add the link, then

brodybits:
@janpio can you take one more look at this?

I think we need to make this update. Thanks!

brodybits:
Thanks @janpio for the quick review!

PR #377, opened 2020-05-13
breaking: use @cordova/eslint-config w/ applied fix

erisu:
### Motivation and Context

Use Cordova's ESLint Configs.

https://github.com/apache/cordova/issues/142

Requires: #376

### Testing

- `npm t`

### Checklist


.eslintrc.yml
```diff
@@ -1,10 +1,30 @@
...
+  - files: [spec/**/*.js]
+    extends: '@cordova/eslint-config/node-tests'
+
+  - files: [cordova-js-src/**/*.js]
+    extends: '@cordova/eslint-config/browser'
+
+    globals:
+      cordova: readonly
+      WinJS: readonly
+      Windows: readonly
```
timbru31, line 30: EOL is missing 😇

template/cordova/prebuild.js
```diff
@@ -30,20 +30,20 @@ module.exports = function patch (platform) {
     var url = require('url');
 
     var basejsSrcMap = {
-        '10': '/www/WinJS/js/base.js',
```
timbru31, line 33: Above we've had `'10.0'` - which is correct?

package.json
```diff
@@ -17,11 +17,11 @@
     "apache"
   ],
   "scripts": {
-    "test": "npm run eslint && npm run test-unit && npm run test-e2e",
+    "test": "npm run lint && npm run test-unit && npm run test-e2e",
     "test-unit": "jasmine",
     "test-e2e": "jasmine spec/e2e/endtoend.spec.js",
     "cover": "nyc jasmine",
-    "eslint": "eslint . \"+(bin|template)/**/!(*.*)\""
+    "lint": "eslint . \"+(bin|template)/**/!(*.*|gitignore)\""
```
timbru31, line 24: Same. Can't we use `eslint .`?

template/cordova/prebuild.js
erisu: Not sure. I don't use Windows.

And reverted the change. Its fixed now.

package.json
erisu: if I remember correctly, its because some of the files are not JS files but are written with JS.. For example `check_all`, `check_req`... Written this will will force them to be checked as well.

timbru31 (approved):

PR #378, opened 2020-05-13
chore: upgrade & set fs-extra as release dependency

erisu:
### Motivation and Context

- Upgrade `fs-extra` dependency to use latest
- `fs-extra` should be a release dependency, not a dev dependency

Requires: #376

### Testing

- `npm run cover`

### Checklist


timbru31:
For future references, changelog is here: https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md#900--2020-03-19

